PR #9, opened 2020-06-15
Add header guards to header files that lacks them

thomasp85:
Some files were missing header guards making it impossible to include them in multiple files


jimhester:
Thanks!

PR #28, opened 2020-06-27
GHA: test with old Windows toolchain

jeroen:
Given that you are doing compiler specific things, and you want to support old versions of R, I think it is important to also test a configuration with the old Windows toolchain.

I think testing for `r: 3.6` suffices because Windows uses the same toolchain as R 3.3 - 3.6.


jimhester:
Thanks a ton!

PR #40, opened 2020-07-02
Optionally register quietly

richfitz:
This PR adds an argument `quiet` that prevents the informational output from cli during registration. Ideally this would also be passed down by pkgbuild, though this looks like a long call path (build -> build_setup -> update_registration -> compile_rcpp_attributes -> cpp11::cpp_register).

The context here is that we're looking at using cpp11 rather than Rcpp in a code generation system, so am looking to avoid this noise when a user has passed a verbose/quiet flag in at our end - we can use `suppressMessages` but that's a big hammer.

### DIFF
--- a/R/register.R
+++ b/R/register.R
@@ -7,9 +7,10 @@
 #' In order to use `cpp_register()` the `cli`, `decor`, `desc`, `glue`,
 #' `tibble` and `vctrs` packages must also be installed.
 #' @param path The path to the package root directory
+#' @param quiet If `TRUE` suppresses output from this function
 #' @return The paths to the generated R and C++ source files (in that order).
 #' @export
-cpp_register <- function(path = ".") {
+cpp_register <- function(path = ".", quiet = FALSE) {
   stop_unless_installed(c("brio", "cli", "decor", "desc", "glue", "tibble", "vctrs"))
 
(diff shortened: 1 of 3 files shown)


jimhester:
Thanks for working on this!

I think in general we are moving away from explicit `quiet` arguments in functions in favor of classed conditions and (if needed) coarse grained whole package options.

https://github.com/r-lib/cli/issues/156 adds a `cliMessage` class to all cli conditions, so you can then use `suppressMessages(, class = "cliMessage")` in recent versions for R to suppress only the cli messages from the call.

richfitz:
I wonder (getting well off topic here) if the class that gets added to the cli message could be something like `c("cliMessage:cpp11", "cliMessage")` which would allow filtering by the package creating it. But then that all falls apart when the next package up wants to suppress messages created by its grandchild or child differently. We probably just need a better logging package 😀

Please close this if you'd rather not chase arguments everywhere, or take it if you think it's fine.

jimhester:
since pkgbuild and friends already use `quiet` and cpp11 will likely have a relatively limited set of R functions I think we should just do the expedient thing and use a `quiet` argument as you propose here.

Thanks again for the PR!

Thanks!

PR #54, opened 2020-07-23
safe[] should support void returning functions

bkietz:
Also inlined `unwind_protect_unwrap_*`

PR #62, opened 2020-07-28
remove unwind_protect_sexp

bkietz:
The explicit suffix is not necessary since `unwind_protect` is overloaded to handle both return types

https://github.com/r-lib/cpp11/pull/54#discussion_r461597780


jimhester:
👍 thanks

PR #70, opened 2020-08-03
support return types other than void,SEXP

### LINKED ISSUE #67: unwind_protect and safe[] can only be used for SEXP or void return
This introduces unnecessary boilerplate when using a function like `Rf_translateCharUTF8` which returns `const char*`:

https://github.com/r-lib/cpp11/blob/030f91dd0ffedb82eb89f0a6c59d712d1766099f/inst/include/cpp11/as.hpp#L134-L136

It would be fairly straightforward to extend these to allow:
```c++
      return unwind_protect([&] { return Rf_translateCharUTF8(STRING_ELT(from, 0)); })[0];
```

Or even (provided `STRING_ELT` doesn't require protection):
```c++
      return safe[Rf_translateCharUTF8](STRING_ELT(from, 0))[0];
```

bkietz:
fixes #67

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # cpp11 (development version)
 
+* `safe[]` can now work with functions that return any type (#70, @bkietz)
+
 * cpp11 is now able to compile on gcc 4.8.5 (#69, @bkietz)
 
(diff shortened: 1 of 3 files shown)


jimhester:
Thanks!

PR #75, opened 2020-08-05
decay types in wrapping code, not in enable if

jimhester:
This simplifies the enable if logic and should help improve error
messages

### DIFF
--- a/cpp11test/src/test-as.cpp
+++ b/cpp11test/src/test-as.cpp
@@ -25,7 +25,7 @@ context("as_cpp-C++") {
     auto x5 = cpp11::as_cpp<unsigned long>(r);
     expect_true(x5 == 42UL);
 
-    auto x6 = cpp11::as_cpp<const int>(r);
+    auto x6 = cpp11::as_cpp<cpp11::decay_t<const int>>(r);
     expect_true(x6 == 42UL);
 
     UNPROTECT(1);
@@ -103,7 +103,7 @@ context("as_cpp-C++") {
     auto x3 = cpp11::as_cpp<long double>(r);
     expect_true(x3 == 1.2);
 
-    auto x4 = cpp11::as_cpp<const double>(r);
+    auto x4 = cpp11::as_cpp<cpp11::decay_t<const double>>(r);
     expect_true(x4 == 1.2);
 
     UNPROTECT(1);
@@ -150,7 +150,7 @@ context("as_cpp-C++") {
     auto x1 = cpp11::as_cpp<bool>(r);
     expect_true(x1);
 
-    auto x2 = cpp11::as_cpp<const bool>(r);
+    auto x2 = cpp11::as_cpp<cpp11::decay_t<const bool>>(r);
     expect_true(x2);
 
     LOGICAL(r)[0] = FALSE;
@@ -167,7 +167,7 @@ context("as_cpp-C++") {
     auto x1 = cpp11::as_cpp<char>(r);
     expect_true(x1 == 'f');
 
-    auto x2 = cpp11::as_cpp<const char>(r);
+    auto x2 = cpp11::as_cpp<cpp11::decay_t<const char>>(r);
     expect_true(x2 == 'f');
 
     UNPROTECT(1);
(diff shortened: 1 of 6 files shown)


jimhester:
@bkietz if you could look this over, I think it is what you had in mind.

cpp11test/src/test-as.cpp
```diff
@@ -285,7 +285,7 @@ context("as_cpp-C++") {
     expect_true(x2[1] == 2.);
     expect_true(x2[2] == 3.);
 
-    auto x3 = cpp11::as_cpp<const cpp11::doubles&>(r);
+    auto x3 = cpp11::as_cpp<cpp11::decay_t<const cpp11::doubles&>>(r);
```
bkietz, line 288: I'd just delete these test cases; no human should write `as_cpp<non_decayed>`

cpp11test/src/test-as.cpp
```diff
@@ -476,9 +476,10 @@ context("as_cpp-C++") {
     UNPROTECT(1);
   }
 
-  test_that("as_sexp(r_vector<cpp11::r_string>)") {
-    SEXP s1 =
-        PROTECT(cpp11::as_sexp(std::vector<cpp11::r_string>({"foo", "bar", "baz"})));
+  test_that("as_sexp(cpp11::strings)") {
+    cpp11::writable::strings s({"foo", "bar", "baz"});
+    cpp11::strings s0(s);
+    SEXP s1 = PROTECT(cpp11::as_sexp(s0));
```
bkietz, line 482: this isn't transparent to me; did compilation break or is it just a preferred spelling of this test or ...?

inst/include/cpp11/as.hpp
```diff
@@ -174,14 +173,6 @@ enable_if_std_string<T, T> as_cpp(SEXP from) {
   return {as_cpp<const char*>(from)};
 }
```
bkietz, line 175:
```suggestion
/// If your compilation fails here you are calling as_cpp<T> with a T
/// which is not decayed; consider rewriting or wrapping T into cpp11::decay_t<T>.
template <typename T>
enable_if_t<!std::is_same<decay_t<T>, T>::value, T> as_cpp(SEXP from) = delete;
```

cpp11test/src/test-as.cpp
jimhester: yeah, I guess there is little point in keeping them

cpp11test/src/test-as.cpp
jimhester: Yeah this particular case `as_sexp(cpp11::string)` is something that occurs in [vroom](https://github.com/r-lib/vroom) but we didn't use it in the cpp11test code at all, so I added a test case for it specifically.

bkietz:
@jimhester re `unmove`: this *will* be necessary if we want to support registering functions with parameters which are mutable references. I think it'd be better not to support these however since the connotation of a mutable reference parameter is a modifiable object from an enclosing scope, whereas we'll always be providing a temporary (and unmove just allows us to wrap that temporary in an ostensibly mutable reference). Specifically, if a user registers

```c++
void append_1(std::vector<std::string>& char_vec) {
  char_vec.push_back("one");
}
```

they might reasonably expect that to mutate a character vector in R the same way calling it from other C++ code would.

cpp11test/src/test-as.cpp
bkietz: alright, but let's keep the original one as well
```suggestion
  test_that("as_sexp(cpp11::strings)") {
    cpp11::writable::strings s({"foo", "bar", "baz"});
    cpp11::strings s0(s);
    SEXP s1 = PROTECT(cpp11::as_sexp(s0));
    
    SEXP s2 = PROTECT(cpp11::as_sexp(std::vector<cpp11::r_string>({"foo", "bar", "baz"})));
```

bkietz:
@jimhester looks like there are a few remaining test cases which call `as_cpp<non decayed>` https://github.com/r-lib/cpp11/pull/75/checks?check_run_id=950884317#step:7:122

bkietz:
@jimhester these changes seem to have been merged but the PR is still open; I think you made the changes to r-lib/master locally and pushed?

jimhester:
yeah, I'll fix it soon

PR #78, opened 2020-08-06
check for compilation errors with GCC 4.8

### LINKED ISSUE #77: gcc 4.8 validation in CI
To ensure that c++11 features are not used which might break gcc 4.8, it'd be useful to have a CI job which builds using that compiler

bkietz:
The stock compiler on centos 7 is gcc 4.8, so using that OS should verify nothing breaks under that compiler.

closes #77

Note: currently includes `gcc48_error()` to ensure that CI catches a known gcc 4.8 glitch. This will of course be removed


jimhester:
I can get this setup this morning

bkietz:
Looks like we'll need a docker action; centos-7 is not in the set of github-hosted virtual environments

.github/workflows/R-CMD-check.yaml
```diff
@@ -73,6 +73,12 @@ jobs:
           sudo apt install g++-6 -y
           echo $'CXX1X=g++-6\nCXX11=g++-6' >> ~/.R/Makevars
 
+      - name : Install compiler version
+        if: matrix.config.custom == 'gcc 4.8'
+        run: |
+          apt-get install -y g++-4.8
+          echo $'CXX1X=g++-4.8\nCXX11=g++-4.8' >> ~/.R/Makevars
```
bkietz, line 80: @jimhester `CXX1X` doesn't look right; did you mean to set the C compiler as well?

jimhester: No, it is correct `CXX1X` is what R used for C++11 on older R versions

jimhester: Not that we are using an older R version in this particular case, but mainly there for consistency's sake.

jimhester:
Ok, after [`a216c45` (#78)](https://github.com/r-lib/cpp11/pull/78/commits/a216c450fdea82d7719e2032fd98c38fd0d8f711) all the CI builds are passing *except* for the 4.8 one due to the known glitch. I will delete that code and merge this.

bkietz:
:+1:

PR #79, opened 2020-08-06
add a workflow to check clang-format

bkietz:
Adds a convenience `make format` target to call clang-format on `**/*.hpp`. Checks for clang-format-10 specifically to avoid the subtle differences between clang-format versions.

Concerns:
- ~~I'd like to run this on `**/*.cpp` as well, but `cpp11test/src/cpp11.cpp` is generated and I'm not sure formatting it is appropriate. Also `cpp11test/R/cpp11-bindings.cpp` would match that glob and is also generated but it seems it *should* be named `cpp11test/R/cpp11-bindings.R`?~~ Generated code is marked with `// clang-format off`
- I've used `make` conditionals and `find` to implement globbing, version checking, etc. It might be preferable to move that logic to an R script
- We'd probably like to add the same functionality for the R source using `formatR` or similar


jimhester:
Should this use run-clang-format.py? (https://github.com/Sarcasm/run-clang-format#how-to-use) it supports a `.clang-format-ignore` file to exclude files.

Alternatively (and maybe a better suggestion) is maybe we should just add a `// clang-format off` comment in the generated code

jimhester:
I think I will add `//clang format off` to the generated code, it is the simplest solution.

.github/workflows/format.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Install ClangFormat
+        run: sudo apt-get install -y clang-format-10
+
+      - name: Run ClangFormat
+        run: make format clang_format=clang-format-10
+
+      - name: Check for a non-empty diff
+        run: |
+          if [ "$(git diff --name-only)" != "" ]
```
jimhester, line 24: I have used `git diff-files -U --exit-code` for this in the past (https://github.com/tidyverse/readr/blob/master/travis/check_format.sh#L1-L13)

I think it is a bit safer than checking for empty names, though probably what you have is fine in practice

jimhester:
Looks good, even picked up an issue :)

I will merge this and fix up the formatting issue after the merge

Thanks!

PR #81, opened 2020-08-06
Avoid c++ in unwind protect

jimhester:


jimhester:
Thanks both for reviewing!

PR #92, opened 2020-08-15
Preserved list refactor

bkietz:
- Renamed from `protect_sexp/release_protect/...` to `preserved.insert/preserved.release` to highlight the similarity in intent to `R_PreserveObject` and decrease the semantic overload of "protect"
- Moved the doubly-linked list of preserved objects from a static object in c++ named `cpp11::protect_list` to an internal in R named `cpp11:::.preserve_list`. This allows the list to be a singleton since any shared object we load will look up the same `cpp11:::.preserve_list`. By contrast a new c++ static variable will be instantiated for each translation unit we compile, resulting in multiple unrelated preserve_lists:
  ```r
  cpp11::cpp_function('SEXP protect_one(SEXP e) { cpp11::protect_sexp(e); return e; }')
  cpp11::cpp_function('int protected_count() { return Rf_length(cpp11::protect_list) - 1; }')
  for (i in 1:10) protect_one(i)
  protected_count() # 0, since it comes from a different translation unit from protect_one
  ```
- Inlined `cpp11::preserve()`
- `sexp& sexp::operator=(const sexp& rhs)` did not release its data before acquiring a new preservation token

### DIFF
--- a/inst/include/cpp11/list.hpp
+++ b/inst/include/cpp11/list.hpp
@@ -5,7 +5,7 @@
 #include "cpp11/R.hpp"                // for SEXP, SEXPREC, SET_VECTOR_ELT
 #include "cpp11/attribute_proxy.hpp"  // for attribute_proxy
 #include "cpp11/named_arg.hpp"        // for named_arg
-#include "cpp11/protect.hpp"          // for protect_sexp, release_protect
+#include "cpp11/protect.hpp"          // for protect_sexp, preserved.release
 #include "cpp11/r_string.hpp"         // for r_string
 #include "cpp11/r_vector.hpp"         // for r_vector, r_vector<>::proxy
 #include "cpp11/sexp.hpp"             // for sexp
@@ -75,7 +75,7 @@ template <>
 inline r_vector<SEXP>::r_vector(std::initializer_list<SEXP> il)
     : cpp11::r_vector<SEXP>(safe[Rf_allocVector](VECSXP, il.size())),
       capacity_(il.size()) {
-  protect_ = protect_sexp(data_);
+  protect_ = preserved.insert(data_);
   auto it = il.begin();
   for (R_xlen_t i = 0; i < capacity_; ++i, ++it) {
     SET_VECTOR_ELT(data_, i, *it);
@@ -86,7 +86,7 @@ template <>
 inline r_vector<SEXP>::r_vector(std::initializer_list<named_arg> il)
     : cpp11::r_vector<SEXP>(safe[Rf_allocVector](VECSXP, il.size())),
       capacity_(il.size()) {
-  protect_ = protect_sexp(data_);
+  protect_ = preserved.insert(data_);
   int n_protected = 0;
 
   try {
@@ -102,7 +102,7 @@ inline r_vector<SEXP>::r_vector(std::initializer_list<named_arg> il)
       UNPROTECT(n_protected);
     });
   } catch (const unwind_exception& e) {
-    release_protect(protect_);
+    preserved.release(protect_);
     UNPROTECT(n_protected);
     throw e;
   }
@@ -114,8 +114,8 @@ inline void r_vector<SEXP>::reserve(R_xlen_t new_capacity) {
                               : safe[Rf_xlengthgets](data_, new_capacity);
 
   SEXP old_protect = protect_;
-  protect_ = protect_sexp(data_);
-  release_protect(old_protect);
+  protect_ = preserved.insert(data_);
+  preserved.release(old_protect);
 
   capacity_ = new_capacity;
 }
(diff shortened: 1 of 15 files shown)


bkietz:
"testing if installed package can be loaded from temporary location" failed with `cpp11:::.preserve_list` because the cpp11 namespace doesn't exist for that step. My first approach to resolve this was to try to create that namespace from c++ when it isn't found. I can't figure out how to do that, so for now I just dump `.cpp11_preserve_list` into the global env. There's probably a better way

jimhester:
One issue with this approach and part of the reason I didn't mind having multiple protect lists for each compilation unit is currently there is no requirement for the cpp11 package itself to be loaded.

An alternative would be to create the protect list in the init function in the generated code. This would give us one protect list per loaded R package that uses cpp11.

https://github.com/r-lib/cpp11/blob/07b7ba7dee2da8c71b1c9fc35c487a6f7b286243/cpp11test/src/cpp11.cpp#L382

bkietz:
With multiple protect lists, `release_existing_protections()` only applies to some protections. If that's intentional then we should document the subtlety; in interactive use one might write
```r
cpp11::cpp_function('int release_all() {
  int n = Rf_length(cpp11::protect_list) - 1;
  cpp11::release_existing_protections();
  return n;
}')
```
which wouldn't have the desired effect.

jimhester:
Yeah, though code needing to manually call `release_existing_protections()` should be pretty rare in practice, it is only needed on older R versions.

Creating the list as an object in the actual `R_GlobalEnv` is likely against CRAN policies, and as mentioned the cpp11 package is not guaranteed to be loaded (and usually won't be) by dependent packages, so we can't put it in the cpp11 package namespace environment. I think the best we could do is create it in the initialization code, which will give us one protection list for each shared library that uses cpp11.

jimhester:
I just thought of a way to do this that will give us a singleton and is not against CRAN policies.

Setting package specific global options are allowed by CRAN, so we could store the preserve list there.

I did this in https://github.com/r-lib/cpp11/pull/92/commits/9c5dd1520e107c40b397c590c3db11bb9e3a2017

jimhester:
Unfortunately there seems to be something wrong with the implementation, we are getting an error on the CI with R 3.4 (https://github.com/r-lib/cpp11/pull/92/checks?check_run_id=1086545888#step:15:15)

> protect(): protection stack overflow

And I see the same error with some of the tests in [vroom](https://github.com/r-lib/vroom) with this PR.

I am still trying to track down the issue.

jimhester:
Ok so the issue was due to option values being duplicated by the `options()` function. Putting the preserve list in an environment avoids this issue and should resolve the errors.

jimhester (approved):

jimhester:
I think this is all working now, @bkietz if you have time to review my changes I can merge this.

inst/include/cpp11/list.hpp
```diff
@@ -5,7 +5,7 @@
 #include "cpp11/R.hpp"                // for SEXP, SEXPREC, SET_VECTOR_ELT
 #include "cpp11/attribute_proxy.hpp"  // for attribute_proxy
 #include "cpp11/named_arg.hpp"        // for named_arg
-#include "cpp11/protect.hpp"          // for protect_sexp, release_protect
+#include "cpp11/protect.hpp"          // for protect_sexp, preserved.release
```
bkietz, line 8:
```suggestion
#include "cpp11/protect.hpp"          // for preserved
```

inst/include/cpp11/raws.hpp
```diff
@@ -8,7 +8,7 @@
 #include "cpp11/R.hpp"                // for RAW, SEXP, SEXPREC, Rf_allocVector
 #include "cpp11/attribute_proxy.hpp"  // for attribute_proxy
 #include "cpp11/named_arg.hpp"        // for named_arg
-#include "cpp11/protect.hpp"          // for protect_sexp, release_protect
+#include "cpp11/protect.hpp"          // for protect_sexp, preserved.release
```
bkietz, line 11:
```suggestion
#include "cpp11/protect.hpp"          // for preserved
```

inst/include/cpp11/strings.hpp
```diff
@@ -7,7 +7,7 @@
 #include "cpp11/as.hpp"               // for as_sexp
 #include "cpp11/attribute_proxy.hpp"  // for attribute_proxy
 #include "cpp11/named_arg.hpp"        // for named_arg
-#include "cpp11/protect.hpp"          // for protect_sexp, unwind_protect
+#include "cpp11/protect.hpp"          // for preserved.insert, unwind_protect
```
bkietz, line 10:
```suggestion
#include "cpp11/protect.hpp"          // for preserved
```

PR #104, opened 2020-09-17
Refactor conversion to use extensible trait structs

bkietz:
Still a little half baked:
- [x] Move all conversion logic to an extensible trait structs
- [x] Generalize container conversion
- [ ] Fix conversion from altrep vectors
- [x] Add doc indicating how and when to specialize the trait structs

### DIFF
--- a/cpp11test/src/test-as.cpp
+++ b/cpp11test/src/test-as.cpp
@@ -8,6 +8,24 @@
 
 #include "Rcpp.h"
 
+struct Vec3 {
+  double x, y, z;
+};
+
+namespace cpp11 {
+template <>
+struct as_cpp_impl<Vec3> {
+  static Vec3 convert(SEXP from) {
+    if (Rf_isReal(from)) {
+      if (Rf_xlength(from) == 3) {
+        return {REAL_ELT(from, 0), REAL_ELT(from, 1), REAL_ELT(from, 2)};
+      }
+    }
+    stop("Expected three double values");
+  }
+};
+}  // namespace cpp11
+
 context("as_cpp-C++") {
   test_that("as_cpp<integer>(INTSEXP)") {
     SEXP r = PROTECT(Rf_allocVector(INTSXP, 1));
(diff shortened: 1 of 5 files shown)


bkietz:
Odd: https://github.com/r-lib/cpp11/pull/104/checks?check_run_id=1130882732#step:12:198 what feature macro should I use to disable that test case? Also, why is the `install cpp11 and cpp11test` step marked successful (the failure is actually raised when trying to run the unbuilt tests)

jimhester:
You can use `#if defined(__APPLE__) && defined(R_VERSION) && R_VERSION >= R_Version(3, 5, 0)` to guard `R_compact_intrange`. ALTREP is only available on R 3.5.0+ and that symbol is hidden on linux and windows, but not on macOS, which is why we guard for apple.

`utils::install.packages()` the installation function shipped with base R only issues warnings when a package installation fails, not a true error.

d9eea11 should convert those warnings to failures and then fail the build in the cpp11test install step in the future.

cpp11test/src/test-as.cpp
```diff
@@ -209,6 +239,25 @@ context("as_cpp-C++") {
     UNPROTECT(1);
   }
 
+#if defined(__APPLE__) && defined(R_VERSION) && R_VERSION >= R_Version(3, 5, 0)
+  test_that("as_cpp<std::vector>(ALTREP_INTSXP)") {
+    SEXP r = PROTECT(R_compact_intrange(1, 5));
+
+    expect_true(ALTREP(r));
+    auto x1 = cpp11::as_cpp<std::vector<int>>(r);
+    expect_true(ALTREP(r));
```
bkietz, line 248: @jimhester  I really expected *something* here to fail since `as_cpp<IntegerContainer>()` currently relies on `INTEGER()`, how is that working with ALTREP?

jimhester: The vector is still an altrep when you call `INTEGER()` on it, but it is then a _materialized_ altrep vector.

The way to test that you are not materializing an alterep vector is to check that `R_altrep_data2(x) == R_NilValue`

jimhester: e.g. after doing this the second test fails, as you were expecting it to.

```
diff --git a/cpp11test/src/test-as.cpp b/cpp11test/src/test-as.cpp
index 6a3282c..0476ef2 100644
--- a/cpp11test/src/test-as.cpp
+++ b/cpp11test/src/test-as.cpp
@@ -243,9 +243,9 @@ context("as_cpp-C++") {
   test_that("as_cpp<std::vector>(ALTREP_INTSXP)") {
     SEXP r = PROTECT(R_compact_intrange(1, 5));
 
-    expect_true(ALTREP(r));
+    expect_true(R_altrep_data2(r) == R_NilValue);
     auto x1 = cpp11::as_cpp<std::vector<int>>(r);
-    expect_true(ALTREP(r));
+    expect_true(R_altrep_data2(r) == R_NilValue);
 
     expect_true(x1.size() == 5);
     int expected = 1;
```

PR #114, opened 2020-10-09
Mark r_vector::const_iterator::operator* as const

### LINKED ISSUE #113: r_vector::const_iterator::operator* is not const
This means `cpp11::r_vector` can break otherwise valid generic code, since `LegacyInputIterator` is required to be const dereferencable.

```r
cpp_function('void fail(cpp11::integers ints) {
  const auto it = ints.begin();
  *it;
}', quiet = FALSE)
#> error: passing 'const cpp11::r_vector<int>::const_iterator' as 'this' argument discards qualifiers [-fpermissive]
#>     return *it;
#>             ^~
```

xhochy:
Fixes #113


xhochy:
Nicely, these were all fixes that were needed to get Arrow R building on Windows using clang.

xhochy:
cc @bkietz Any ideas how to fix this properly?

jimhester:
Thanks for working on this!

Issue #116, opened 2020-10-13
attempting to save R options causes C stack overflow


kevinushey:
Sorry, but I'm not able to generate a `reprex` right now (it's failing with errors I don't recognize). Here's an otherwise reproducible example:

```
library(readr)

writeLines("a,1", con = "/tmp/test.csv")
readr::read_csv("/tmp/test.csv")

opts <- options()
save(opts, file = "/tmp/options.Rdata")
```

I see:

```
> save(opts, file = "/tmp/options.Rdata")
Error: C stack usage  7969392 is too close to the limit
```

This appears to be due to something funky in the `cpp11_preserve_env` object. (I can't inspect the `cpp11_preserve_list` object without seeing weird R errors)

jimhester:
Fixed by https://github.com/r-lib/cpp11/pull/117

PR #117, opened 2020-10-14
Store the protect list as an XPtr rather than in an environment

jimhester:
Using an environment as previously done generates a stack overflow when
serialized.

The data part of XPtrs is not serialized, so avoids this issue, while
also avoiding the copying issues that prompted use of an environment.


jimhester:
Fixes #116

PR #122, opened 2020-10-28
Make preserve list more robust to bad values

### LINKED ISSUE #121: handle case where cpp11_preserve_xptr is a null pointer
RStudio attempts to save and load options after session restarts (e.g. after installing and reloading an R package), leading to this state:

```
> getOption("cpp11_preserve_xptr")
<pointer: 0x7faf18c1c450>

Restarting R session...

Using libraries at paths:
- /Users/kevinushey/Library/R/4.0/library
- /Library/Frameworks/R.framework/Versions/4.0/Resources/library
> library(euclid)
> getOption("cpp11_preserve_xptr")
<pointer: 0x0>
```

This can unfortunately cause issues for packages which try to use `cpp11`, as (if I understand correctly) `cpp11` assumes that this is a valid + live pointer. For example:

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x28)
  * frame #0: 0x00000001094b4554 libR.dylib`CDR(e=0x0000000000000000) at memory.c:4124:29 [opt]
    frame #1: 0x000000010d9f3f30 euclid.so`cpp11::$_0::insert(this=<unavailable>, obj=0x00007fe5c10803e8) at protect.hpp:227:17 [opt]
    frame #2: 0x000000010d9f3e65 euclid.so`cpp11::r_vector<double>::r_vector(this=0x00007ffee8285148, data=0x00007fe5c10803e8) at r_vector.hpp:362:26 [opt]
    frame #3: 0x000000010d9d1c20 euclid.so`::_euclid_create_exact_numeric(SEXP) [inlined] cpp11::r_vector<double>::r_vector(this=<unavailable>, data=<unavailable>) at r_vector.hpp:365:33 [opt]
    frame #4: 0x000000010d9d1c18 euclid.so`::_euclid_create_exact_numeric(SEXP) [inlined] std::__1::enable_if<((!(is_smart_ptr<cpp11::r_vector<double> >::value)) && (std::is_class<cpp11::r_vector<double> >::value)) && (std::is_constructible<cpp11::r_vector<double>, SEXPREC*>::value), cpp11::r_vector<double> >::type cpp11::as_cpp<cpp11::r_vector<double> >(from=<unavailable>) at as.hpp:82 [opt]
    frame #5: 0x000000010d9d1c18 euclid.so`::_euclid_create_exact_numeric(numeric=<unavailable>) at cpp11.cpp:403 [opt]
    frame #6: 0x000000010942161c libR.dylib`R_doDotCall(ofun=<unavailable>, nargs=<unavailable>, cargs=0x00007ffee8287620, call=0x00007fe5f1311598) at dotcode.c:598:17 [opt]
< ... >
```

```
(lldb) f 1
frame #1: 0x000000010d9f3f30 euclid.so`cpp11::$_0::insert(this=<unavailable>, obj=0x00007fe5c10803e8) at protect.hpp:227:17 [opt]
   224      PROTECT(obj);
   225
   226      // Add a new cell that points to the previous end.
-> 227      SEXP cell = PROTECT(Rf_cons(list_, CDR(list_)));
   228
   229      SET_TAG(cell, obj);
   230
(lldb) f 2
frame #2: 0x000000010d9f3e65 euclid.so`cpp11::r_vector<double>::r_vector(this=0x00007ffee8285148, data=0x00007fe5c10803e8) at r_vector.hpp:362:26 [opt]
   359  template <typename T>
   360  inline r_vector<T>::r_vector(const SEXP data)
   361      : data_(valid_type(data)),
-> 362        protect_(preserved.insert(data)),
   363        is_altrep_(ALTREP(data)),
   364        data_p_(get_p(ALTREP(data), data)),
   365        length_(Rf_xlength(data)) {}
```

Would `cpp11` also be willing to handle this case where `cpp11_preserve_xptr` is not a valid + live external pointer, or should RStudio just avoid serializing + deserializing this pointer?

jimhester:
XPtrs are marked as null when restoring, so we need to check for this
and we also now check that the option is a pairlist before trying to
read its value

Fixes #121

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # cpp11 (development version)
 
+* The preserve list is now more robust to invalid values, such as when the XPtr has no address or if non-xptr's are stored in the option. This fixes errors when reloading packages using cpp11 and RStudio's session restores.
+
 # cpp11 0.2.3
 
 * `r_vector::const_iterator::operator*` is now a const method (#113, @bkietz, @xhochy)
--- a/inst/include/cpp11/protect.hpp
+++ b/inst/include/cpp11/protect.hpp
@@ -223,6 +223,10 @@ static struct {
 
     PROTECT(obj);
 
+    if (TYPEOF(list_) != LISTSXP) {
+      list_ = get_preserve_list();
+    }
+
     // Add a new cell that points to the previous end.
     SEXP cell = PROTECT(Rf_cons(list_, CDR(list_)));
 
@@ -289,9 +293,15 @@ static struct {
   // `CPP11_UNWIND`, so if an error occurs we will not catch the C++ exception
   // that safe emits.
   static void set_option(SEXP name, SEXP value) {
-    SEXP opt = SYMVALUE(Rf_install(".Options"));
+    static SEXP opt = SYMVALUE(Rf_install(".Options"));
     SEXP t = opt;
     while (CDR(t) != R_NilValue) {
+      if (TAG(CDR(t)) == name) {
+        opt = CDR(t);
+        SET_TAG(opt, name);
+        SETCAR(opt, value);
+        return;
+      }
       t = CDR(t);
     }
     SETCDR(t, Rf_allocList(1));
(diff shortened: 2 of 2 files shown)


jimhester:
This refactors the preserve list code to be more robust to bad values. We now verify the types are what we expect, including when using the `list_` value in `preserved.insert()` and explicitly check for the XPtr being null, which happens when the XPtr is serialized and restored.

I also verified that after this PR and recompiling readr with it you can do repeated Install and RStarts within RStudio without crashes.

It would be good to have another set of eyes though, so @kevinushey if you could review it I would appreciate it.

kevinushey (approved):
LGTM!

jimhester:
Thanks!

PR #146, opened 2021-01-14
describe addition of useDynLib in cpp11 vignette

mpadge:
Without this, explicitly following the current instructions of just adding `LinkingTo` fails, likely leaving many to attempt `usethis::use_cpp11()`, which also fails because of [this :bug:](https://github.com/r-lib/usethis/pull/1329). Once `use_cpp11()` works after *only* adding `LinkingTo`, and not `useDynLib`, and once the vignette has been updated with this PR, then the entire procedure should be sufficiently robust that any potential combination of these approaches should give the desired result of enabling use of `cpp11`. Thanks as always @jimhester  for the awesome work. Hope that helps!


jimhester:
Thank you!

PR #157, opened 2021-03-26
Add dir argument to cpp_source

### LINKED ISSUE #156: Make it easier to debugger to do source mapping with cpp11::cpp_source
I have a repo [vscode-rcpp-demo](https://github.com/renkun-ken/vscode-rcpp-demo) to demonstrate how to debug Rcpp C++ code in a package project in VSCode. I also create a similar repo [vscode-cpp11-demo](https://github.com/renkun-ken/vscode-cpp11-demo) to do the same for cpp11. Both work nicely if the project is an R package.

However, in both cases, we cannot debug C++ code with `Rcpp::sourceCpp()` and `cpp11::cpp_source()` unless we configure [source mapping](https://code.visualstudio.com/docs/cpp/launch-json-reference#_sourcefilemap) for the C++ debugger. I raised an issue https://github.com/RcppCore/Rcpp/issues/1041 on Rcpp side, but it seems not trivial to change how the compilation works to support this.

I do some experiment with cpp11, and it seems straightforward to support this: just allow the user to customize https://github.com/r-lib/cpp11/blob/master/R/source.R#L69 so that the directory is pre-determined, and then we can trigger the C++ debugger with the following in a non-package project:

`./code.cpp`:

```cpp
#include <cpp11.hpp>
#include <iostream>
using namespace cpp11;

[[cpp11::register]]
double calc_sum(doubles x)
{
  double sum = 0;
  for (int i = 0; i < x.size(); ++i) {
    std::cout << i << std::endl;
    sum += x[i];
  }
  return sum;
}
```

`./code.R`:

```r
cpp11::cpp_source("code.cpp", tmpdir = "./tmp")
calc_sum(c(1, 2, 3))
```

and create a source mapping in `.vscode/launch.json`:

```json
"sourceFileMap":{
        "${workspaceFolder}/tmp/src/": "${workspaceFolder}/"
      }
```

Then the breakpoints and other debugger features could work correctly with the source mapping between the original source files and the copied and generated source files in `./tmp/src`.

Do you think if it makes sense to allow the user to supply the temp dir for debugging purposes?

renkun-ken:
Closes #156.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # cpp11 0.2.7
 
 * Fix a transient memory leak for functions that return values from `cpp11::unwind_protect()` and `cpp11::safe` (#154)
+* `cpp_source()` now gets an argument `dir` to allow customized temporary directory to store generated source files.
+  It makes it easier to debug C++ source files in non-package project via source mapping. (@renkun-ken, #156)
 
 # cpp11 0.2.6
 
--- a/R/source.R
+++ b/R/source.R
@@ -16,6 +16,8 @@
 #' @param cxx_std The C++ standard to use, the `CXX_STD` make macro is set to
 #'   this value. The default value queries the `CXX_STD` environment variable, or
 #'   uses 'CXX11' if unset.
+#' @param dir The directory to store the generated source files. `tempfile()` is
+#'   used by default. The directory will be removed if `clean` is `TRUE`.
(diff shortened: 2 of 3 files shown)


jimhester:
Thanks! Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jimhester:
Thanks a bunch!

PR #160, opened 2021-03-30
Fix handling of whitespace in Makevars filename

klmr:
This PR adds escaping of spaces in path names that are written to the `Makevars` file used by ‘cpp11’ for compilation.

I came across this issue because I’m keeping my R package library on macOS in the standard config location, which is `~/Library/Application Support` (note the space). Since ‘cpp11’ adds its own installation directory to the include path, this consequently causes an error when invoking e.g. `cpp_source()`, which looks as follows:

>     clang: error: no such file or directory: 'Support/R/4.0/library/cpp11/include'
>     make: *** [/private/var/folders/dr/rp46tjqs5qzfj_ghjpxqrttr0000gn/T/RtmpMtBJTo/file13e521450b665/src/code_13e523ef6ab4e.o] Error 1
>     Error: Compilation failed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: cpp11
 Title: A C++11 Interface for R's C Interface
-Version: 0.2.7
+Version: 0.2.7.9000
 Authors@R: 
     c(person(given = "Jim",
              family = "Hester",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# cpp11 (development version)
+
+* Fix handling of spaces in Makevars include filenames (@klmr, #160)
+
 # cpp11 0.2.7
 
 * Fix a transient memory leak for functions that return values from `cpp11::unwind_protect()` and `cpp11::safe` (#154)
--- a/R/source.R
+++ b/R/source.R
@@ -153,12 +153,13 @@ generate_include_paths <- function(packages) {
     if (is_windows()) {
       path <- shQuote(utils::shortPathName(path))
     }
-    out[[i]] <- paste0("-I", shQuote(path))
+    out[[i]] <- paste0("-I", path)
   }
   out
 }
 
 generate_makevars <- function(includes, cxx_std) {
+  includes <- gsub(" ", "\\\\ ", includes)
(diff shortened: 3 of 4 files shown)


jimhester:
I think I would prefer to handle this by unconditionally surrounding the paths in single quotes at https://github.com/r-lib/cpp11/blob/44b309ac266b48751fe1b8e62f0eae568d352106/R/source.R#L156

Also 
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jimhester:
Looks good except that windows apparently can't handle the single quotes 🙈

We could try double quotes, though then I worry POSIX systems will interpolate variables in them. I guess we might have to live with the backslashes...

klmr:
> We could try double quotes, though then I worry POSIX systems will interpolate variables in them. I guess we might have to live with the backslashes...

What about `shQuote`? It *should* handle this.

jimhester:
Sure, `shQuote` seems like a good idea

klmr:
Hmm. The check failure on Windows (release) seems unrelated (?). But it still fails. :-(

klmr:
I’m not entirely sure what the issue is but it seems that Make on Windows doesn’t like double quotes. So, should I revert back to the first version (escape whitespace with backslashes)?

(If that’s alright I’ll push a new commit rather than amending and force-pushing — when squashing the PR the effect would be the same.)

jimhester:
Yeah, we can revert back to using backslashes and I will squash this before merging.

jimhester:
Also at least on my macOS system `~/Library/ApplicationSupport` _without_ a space is a symlink to `~/Library/Application Support`. I don't think I added this symlink myself, so you might save yourself grief by using that as your library path rather than the one with a space.

klmr:
> you were accidentily calling shQuote twice on windows

Oh wow, I completely missed that. Thanks for spotting that, and for adding the tests.

> Also at least on my macOS system `~/Library/ApplicationSupport` without a space is a symlink to `~/Library/Application Support`.

Yes, that symlink is always there and I fall back to it when I need to — but by default I use the canonical path intentionally to flush out just this kind of issue during development. There are actually very few issues: In my experience, Make is the only modern, widely used tool that doesn’t support spaces.

Issue #143, opened 2021-01-09
.cpp file changes not reflected after successive call to cpp_source


sccmckenzie:
After initial call to `cpp_source('file.cpp')`, any subsequent calls to `cpp_source('file.cpp')` appear to have no effect on the actual compiled function - regardless of changes made to `file.cpp`.

### file.cpp
```cpp
#include "cpp11.hpp"
using namespace cpp11;

[[cpp11::register]]
int one() {
  return 1;
}
```

### R sesssion
```r
library(cpp11)
cpp_source("file.cpp")

one()
#> [1] 1
```

### Make changes to file.cpp (return 2 instead of 1), save...
```cpp
#include "cpp11.hpp"
using namespace cpp11;

[[cpp11::register]]
int one() {
  return 2; // change to 2
}
```

### Back to R session. Return value is still 1.
```r
rm(one)
cpp_source("file.cpp")

one()
#> [1] 1
```

I have tried reinstalling cpp11 and Rtools - same result.

```r
sessioninfo::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value                       
#>  version  R version 4.0.3 (2020-10-10)
#>  os       Windows 10 x64              
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United States.1252  
#>  ctype    English_United States.1252  
#>  tz       America/Chicago             
#>  date     2021-01-08                  
#> 
#> - Packages -------------------------------------------------------------------
#>  package     * version date       lib source        
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.0.2)
#>  cli           2.2.0   2020-11-20 [1] CRAN (R 4.0.3)
#>  cpp11       * 0.2.4   2020-11-05 [1] CRAN (R 4.0.3)
#>  crayon        1.3.4   2017-09-16 [1] CRAN (R 4.0.2)
#>  digest        0.6.25  2020-02-23 [1] CRAN (R 4.0.2)
#>  evaluate      0.14    2019-05-28 [1] CRAN (R 4.0.2)
#>  fansi         0.4.1   2020-01-08 [1] CRAN (R 4.0.2)
#>  glue          1.4.2   2020-08-27 [1] CRAN (R 4.0.2)
#>  highr         0.8     2019-03-20 [1] CRAN (R 4.0.2)
#>  htmltools     0.5.0   2020-06-16 [1] CRAN (R 4.0.2)
[...]

PR #176, opened 2021-06-06
Don't modify the file name on cpp_source, only the shared lib name

vspinu:
Modified base names break navigation to  the original file location on errors (in my case emacs/ESS). Modifying only the shared lib path at the end would have the same effect of registering dyn lib with a new name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ vignettes/*_cache
 compile_commands.json
 revdep/*/
 talk/
+TAGS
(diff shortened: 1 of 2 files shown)


jimhester:
I think the shared library passed to `dyn.load()` needs to have a unique name on Windows or else it is not loaded by R. 

I believe this proposed change would break https://github.com/r-lib/cpp11/issues/143

vspinu:
In my PR dll names are changed, the file name is not. So dyn.load() still receives different dll each time. 

I might not have been very clear, the errors I am referring to are compiler errors which refer to the temporary cpp files, what happens to the dll latter does not matter.

jimhester:
Oh I see what you are doing now, didn't quite grasp it the first time I read your proposed changes.

Thanks for the clarification, I think it looks OK.

PR #178, opened 2021-06-07
Add cpp11 faq documentation

sbearrows:
Initial pull request draft. Let's work on this together!


vignettes/cpp11_faq.Rmd
```diff
@@ -0,0 +1,184 @@
...
+)
+```
+
+Intended to answer common questions about the usage of cpp11 objects
+
+## FAQ
+
+* How do I modify a vector in place?
+
+  * Using `writable::` will always make a copy, but it has a move constructor, so you can use `cpp11::writable::integers(std::move(x))` and it won't make a copy of the data.
```
jimhester, line 23: I think we should move this one down, I generally think doing this is a mistake many of the times, so don't want to make it too prominent.

vignettes/cpp11_faq.Rmd
```diff
@@ -0,0 +1,184 @@
...
+
+* How do I add elements to a named list?
```
jimhester, line 25: I think we should explain what we are doing first, then show the code, not just showing code.

Maybe something like

> Use the vector's `push_back()` method along with named literals.

In the code examples you can use the cpp11 knitr engine, you can do this with the following header.

    ```{cpp11}

For example https://github.com/r-lib/cpp11/blob/4ef13b7ecdcb9cbd058456b7a18c1bb3ba665d05/vignettes/motivations.Rmd#L134-L144

vignettes/cpp11_faq.Rmd
```diff
@@ -0,0 +1,184 @@
...
+
+    `cpp11::writable::list x;`
+
+* How do I retrieve (named) elements from a named vector/list?
+
+    `x["foo"]`
+
+* How can I tell whether a vector is named?
+
+    `x.names() == R_NilValue;`
```
jimhester, line 96: I think we should also open an issue to add a `named()` function, for the same reasons as #182

jimhester (changes requested):
We should number the FAQ entries, and I would put each question in a header, e.g. `####` or `###` so they they are linkable.

vignettes/cpp11_faq.Rmd
sbearrows: Sounds good!

vignettes/cpp11_faq.Rmd
sbearrows: That's a great idea! I'll open an issue for this.

vignettes/cpp11_faq.Rmd
```diff
@@ -0,0 +1,183 @@
...
+  my_env["foo"] == 1;
+
+  my_env["bar"] == R_UnboundValue;
+  ```
+
+* How can I create a `cpp11:raws` from a `std::string`?
+
+  * Push back the contents individually
+  ```
+  cpp11::cpp_source(code = '#include <cpp11.hpp>
```
jimhester, line 58: You need to use the cpp11 knitr engine all the places you see `cpp_source()` in these examples.

vignettes/cpp11_faq.Rmd
```diff
@@ -0,0 +1,236 @@
...
+
+`cpp11::writable::list x;`
+
+#### 7. How do I retrieve (named) elements from a named vector/list?
+
+`x["foo"]`
+
+#### 8. How can I tell whether a vector is named?
+
+  TODO
```
jimhester, line 139: We can update this now with `x.named()` right?

sbearrows: Yep! 👍

sbearrows:
I think the fix for #195 caused faq 10 to not work as before so I've changed it a bit but I may be mistaken!

vignettes/cpp11_faq.Rmd
```diff
@@ -170,9 +169,9 @@ cpp11::writable::logicals fn2() {
 }
 ```
     
-  * Otherwise this will return a 0 length logical vector:
```
jimhester: Thanks for updating this, it was a recent change in behavior, but I think we have better behavior now!

Issue #127, opened 2020-11-12
Output a more useful error message if you don't register a function


jimhester:
Right now you get 

```
Error: Can't subset columns that don't exist.
x Columns `name`, `return_type`, and `args` don't exist.
```

Which is not very helpful.

sbearrows:
I tried recreating the error using `devtools::load_all()` in an R package by removing cpp11 decorators but I was still only able to recreate it when the content between `[[]]` was not `cpp11::register` (ex `[[cpp11:register]]`, `[[cpp11::egister]]`, etc). If there are no decorators it doesn't give any error. If there is at least one other function with the correct decorator (so one with the correct decorator and one with the incorrect decorator) the compiler gives an error that is fairly informative. I'll move ahead with the solution we discussed earlier!

PR #180, opened 2021-06-09
Feature add error message for incorrect decorators

sbearrows:
First pass for adding this error message. Error message could prolly still be more informative but let me know what you think!


R/register.R
```diff
@@ -136,6 +136,18 @@ get_registered_functions <- function(decorations, tag, quiet = FALSE) {
   }
 
   out <- decorations[decorations$decoration == tag, ]
+
+  if(any(!decorations$decoration %in% c("cpp11::register", "cpp11::init"))) {
```
jimhester, line 140: maybe store the result in a variable to make this easier to read?

e.g. 

```r
bad_decorations <- !decorations$decoration %in% c("cpp11::register", "cpp11::init")

if (any(bad_decorations)) { # etc..
```

R/register.R
```diff
@@ -136,6 +136,18 @@ get_registered_functions <- function(decorations, tag, quiet = FALSE) {
...
+    lines <- decorations$line[which(!decorations$decoration %in% c("cpp11::register", "cpp11::init"))]
+    if(length(lines) > 1) {
+      stop(call. = FALSE, paste0("Can't capture cpp11 decorators on lines ",
```
jimhester, line 143: Maybe the error should be something like 

```
cpp11 attributes must be either `cpp11::register` or `cpp11::init`:
- Invalid attribute [[cpp11::foobar]] on line 123 in file xyz.
- Invalid attribute [[cpp11::baz]] on line 456 in file abc.
```

R/register.R
sbearrows: Sure!

jimhester:
I think maybe we should pull this check out into a separate `check_valid_attributes()` function, otherwise we would run in twice.

R/register.R
```diff
@@ -136,6 +136,18 @@ get_registered_functions <- function(decorations, tag, quiet = FALSE) {
   }
 
   out <- decorations[decorations$decoration == tag, ]
+
+  if(any(!decorations$decoration %in% c("cpp11::register", "cpp11::init"))) {
+    lines <- decorations$line[which(!decorations$decoration %in% c("cpp11::register", "cpp11::init"))]
```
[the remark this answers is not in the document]
sbearrows: :thumbsup:

R/register.R
sbearrows: Oh that looks great!
Is the best way to make multi-line error messages by using a combination of `paste()` statements with `sep = "\n"` ? Or would this be a `message()` following by an empty `stop()` ?

jimhester: You can actually use [glue](https://github.com/tidyverse/glue) to do this a little more nicely, something like this

```r
bad_lines <- glue::glue_collapse(
  glue::glue("- Invalid attribute `{names}` on line {line} in file '{file}'."),
  "\n"
)

msg <- glue::glue("
  cpp11 attributes must be either `cpp11::register` or `cpp11::init`:
  {bad_lines}
  ")
stop(msg, call. = FALSE)
```

sbearrows: Oh nice!

sbearrows:
fixes #127

Issue #186, opened 2021-06-15
Add `named` method for `r_vector` class


sbearrows:
Add `named()` feature for `r_vector`

Current approach to checking if an `r_vector` is named is not super intuitive.

`named()` should return `x.names() == R_NilValue;`

Similar to issue #182

PR #188, opened 2021-06-23
adding empty to vector classes

### LINKED ISSUE #182: Add `empty` method for `r_vector` classes.
The C++ core guidelines recommend using `empty()` methods where available, because this makes the containers more generic, e.g. you don't have to know that an empty container means the size is 0.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t143-dont-write-unintentionally-non-generic-code

We should add `empty()` methods that return `size() > 0`.

We should probably use `empty()` rather than `is_empty()` to match the `std::vector` APIs, when in doubt I have been trying to stick to that.

sbearrows:
Added news updates for both `named()` and `empty()` to this PR

fixes #182

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # cpp11 (development version)
 
-*Added `named()` and `empty()` methods to r_vector classes, returning true if named/empty (@sbearrows, #182, #186) 
+* Added `x.empty()` method to check if a vector is empty (@sbearrows, #182)
 * New `cpp11::na()` function to return the NA sentinals for R objects(@sbearrows, #179)
 * Incorrectly formatted cpp11 decorators now output a more informative error message (@sbearrows, #127)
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,5 +1,6 @@
 # cpp11 (development version)
 
+*Added `named()` and `empty()` methods to r_vector classes, returning true if named/empty (@sbearrows, #182, #186)
```
jimhester, line 3: We should try to follow https://style.tidyverse.org/news.html#common-patterns when formatting news. Maybe these would be better as two separate entries?

* New `x.named()` method to check if a vector is named (@sbearrrows, #186)
* New `x.empty()` method to check if a vector is empty (@sbearrows, #182)

What do you think?

sbearrows: Sure! Looks good,  and thanks for the links

jimhester:
Hey! Thanks!

PR #189, opened 2021-06-23
added r_vector named() method

### LINKED ISSUE #186: Add `named` method for `r_vector` class

sbearrows:
fixes #186


jimhester:
Thanks a million!

PR #195, opened 2021-06-28
Workaround issue constructing logical vectors of length 1 from 'false'

### LINKED ISSUE #187: `false` not currently working for Windows and Ubuntu platforms
Currently `false` is not working for certain versions of Windows and all versions of Ubuntu on `R CMD check`

jimhester:
Fixes #187

### DIFF
--- a/cpp11test/src/test-logicals.cpp
+++ b/cpp11test/src/test-logicals.cpp
@@ -134,6 +134,16 @@ context("logicals-C++") {
     expect_true(cpp11::is_na(y));
   }
 
+  test_that("FALSE and false") {
+    cpp11::writable::logicals x{FALSE};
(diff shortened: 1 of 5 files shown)


jimhester:
@sbearrows this seems to be sufficient to fix #187. I removed some existing initializer_list<> overloads that no longer seem necessary, as `r_string` objects can be constructed from `const char*` nad `std::string`. These overloads were causing the problem with g++ 5 as the compiler was either picking these instead of the `r_bool` one for `false` for some reason (possibly a compiler bug).

In any case removing these overloads seems to resolve this and the constructor we want is should now be found.

sbearrows (commented):
Looks good!

PR #196, opened 2021-06-29
Add pkgdown site to URL in DESCRIPTION

batpigandme:


jimhester:
I apparently never do this

batpigandme:
You do sometimes! I went through like 180+ packages, so I'd say you've got a decent batting record. 😜

PR #197, opened 2021-06-29
Bug cpp11 decorators

### LINKED ISSUE #193: `linking_to` attribute is no longer supported in `cpp_source()`
This came from a [recent build failure](https://github.com/paleolimbot/wk/runs/2922638631#step:10:84) for a [pkgdown article that uses the cpp11 knitr engine](https://github.com/paleolimbot/wk/blob/master/vignettes/articles/programming.Rmd#L73). I don't think I'm using the attribute incorrectly but it's certainly possible!

Reprex from the `cpp_source()` example page:

``` r
cpp11::cpp_source(
  code = '
#include <cpp11/R.hpp>
#include <RProgress.h>

[[cpp11::linking_to("progress")]]

[[cpp11::register]] void
show_progress() {
  RProgress::RProgress pb("Downloading [:bar] ETA: :eta");

  pb.tick(0);
  for (int i = 0; i < 100; i++) {
    usleep(2.0 / 100 * 1000000);
    pb.tick();
  }
}
')
#> Error: cpp11 attributes must be either `cpp11::register` or `cpp11::init`:
#> - Invalid attribute `cpp11::linking_to` on
#> line 5 in file '/var/folders/bq/2rcjstv90nx1_wrt8d3gqw6m0000gn/T//RtmpH9kS6D/file2d7a258fb142/src/code_2d7a7cbe7751.cpp'.
```

Would it make sense to add `linking_to` to the list of allowed decorations here?

https://github.com/r-lib/cpp11/blob/7a7da573a3bbbfc92c3b0ec6857b5cb57a299478/R/register.R#L282

sbearrows:
Also added new tests for linking_to

fixes #193

### DIFF
--- a/R/register.R
+++ b/R/register.R
@@ -290,7 +290,7 @@ check_valid_attributes <- function(decorations) {
     bad_lines <- glue::glue_collapse(glue::glue("- Invalid attribute `{names}` on
                  line {lines} in file '{file}'."), "\n")
 
-    msg <- glue::glue("cpp11 attributes must be either `cpp11::register`, `cpp11::init` or `cpp11::linking_to`:
+    msg <- glue::glue("cpp11 attributes must be one of `cpp11::register`, `cpp11::init` or `cpp11::linking_to`:
       {bad_lines}
       ")
     stop(msg, call. = FALSE)
--- a/R/register.R
+++ b/R/register.R
@@ -46,6 +46,8 @@ cpp_register <- function(path = ".", quiet = FALSE) {
     return(invisible(character()))
   }
 
+  check_valid_attributes(all_decorations)
+
   funs <- get_registered_functions(all_decorations, "cpp11::register", quiet)
 
   package <- desc::desc_get("Package", file = file.path(path, "DESCRIPTION"))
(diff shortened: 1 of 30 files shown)


R/register.R
```diff
@@ -279,7 +279,9 @@ get_cpp_register_needs <- function() {
 
 check_valid_attributes <- function(decorations) {
 
-  bad_decor <- !decorations$decoration %in% c("cpp11::register", "cpp11::init")
+  bad_decor <- grepl("cpp11::", decorations$decoration) & (!decorations$decoration %in% c("cpp11::register",
```
jimhester, line 282: `startsWith()` would be slightly preferred here.

If you do want to use `grepl()` you need to anchor it to the start of the string, e.g. `grepl("^cpp11::")`

R/register.R
```diff
@@ -288,7 +290,7 @@ check_valid_attributes <- function(decorations) {
     bad_lines <- glue::glue_collapse(glue::glue("- Invalid attribute `{names}` on
                  line {lines} in file '{file}'."), "\n")
 
-    msg <- glue::glue("cpp11 attributes must be either `cpp11::register` or `cpp11::init`:
+    msg <- glue::glue("cpp11 attributes must be either `cpp11::register`, `cpp11::init` or `cpp11::linking_to`:
```
jimhester, line 293:
```suggestion
    msg <- glue::glue("cpp11 attributes must be one of `cpp11::register`, `cpp11::init` or `cpp11::linking_to`:
```

jimhester (commented):
In addition to your changes (which look good!) I think we should remove the `\dontrun{}` block around the example at https://github.com/r-lib/cpp11/blob/532cd3fd9753eca7349f8a9a4c07e85d5d6c440d/R/source.R#L25 and Re-document.

I initially used `\dontrun{}` so users wouldn't be loading a shared library when they ran the examples, but I now think that isn't really a problem and it would have caught this regression during R CMD check.

R/register.R
sbearrows: Yep you're right! Thanks for the review

R/register.R
```diff
@@ -279,7 +279,7 @@ get_cpp_register_needs <- function() {
 
 check_valid_attributes <- function(decorations) {
 
-  bad_decor <- grepl("cpp11::", decorations$decoration) & (!decorations$decoration %in% c("cpp11::register",
+  bad_decor <- startsWith(decorations$decoration, "cpp11::") & (!decorations$decoration %in% c("cpp11::register",
                                                                                          "cpp11::init",
                                                                                          "cpp11::linking_to"))
```
jimhester: Is this still lining up like you wanted?

It might be worth doing the line breaking a bit differently, e.g.

```suggestion
  bad_decor <- startsWith(decorations$decoration, "cpp11::") &
    (!decorations$decoration %in% c("cpp11::register", "cpp11::init", "cpp11::linking_to"))
```

sbearrows: Sure! Looks nice

jimhester:
Thanks!

PR #198, opened 2021-06-30
Use original path names in compiler outputs

### LINKED ISSUE #194: cpp_source reports error with temp file path
`cpp11::cpp_source()` by default uses a temp directory to store the source files to compile. When an error occurs, however, the errors are also reported with a temp path like the following:

```r
> cpp11::cpp_source("indention.cpp")                                                                                                                                                   
/tmp/RtmptYYXie/file3b9a74895bd2a/src/indention.cpp:6:47: error: ‘x’ was not declared in this scope
    6 | SEXP detect_comments(SEXP content, SEXP row = x)
      |                                               ^
make: *** [/usr/lib/R/etc/Makeconf:177: /tmp/RtmptYYXie/file3b9a74895bd2a/src/indention.o] Error 1
make: *** Waiting for unfinished jobs....
Error: Compilation failed.
Backtrace:
1: stop("Compilation failed.", call. = FALSE)
2: cpp11::cpp_source("indention.cpp")
```

where these files are immediately removed. If a code editor (e.g. VS Code) supports detecting file links in the terminal output, user could not jump to the line and column from these paths as they don't exist anymore.

I guess it might make more sense to show the original file paths in the error output.

### LINKED ISSUE #202: Error when calling sourced cpp11 function after multiple `cpp_source` calls
```
cpp11::cpp_source("xyz.cpp")
> boo()
#> [1] 10
> cpp11::cpp_source("xyz.cpp")
> boo()
#> Error in .Call("_xyz_2_boo", PACKAGE = "xyz_2") : 
  "_xyz_2_boo" not available for .Call() for package "xyz_2"
```

This is likely due to our changes to the temp file names. We will have to re-evaluate how we are handling temporary file names.

Restarting the R session removes this error.

jimhester:
This is a first draft at fixing #194.

It works for me locally on macOS. We would need to do some testing on Windows and maybe think about improving the search terms to be more specific. (include the trailing ':') in the error maybe?

Fixes #194
Fixes #202


jimhester:
@sbearrows if you could take this as a starting point and push it further it would be great!

You can use `usethis::pr_fetch(198)` to pull this code down to your machine, and `usethis::pr_push()` to push changes back to this PR after you have commited them.

sbearrows:
Okay definitely!

jimhester:
@sbearrows this looks like it needs a news entry still, and do you think we should try to address #202 in this PR or separately?

sbearrows:
I was thinking I would do #202 in this one for simplicity. Thoughts?

jimhester:
Sure, that is fine!

R/source.R
```diff
@@ -128,16 +123,19 @@ cpp_source <- function(file, code = NULL, env = parent.frame(), clean = TRUE, qu
...
+    # Substitute temporary file path with original file path
+    error_messages <- gsub(file.path(new_path, basename(file)), file.path(orig_path, basename(file)), error_messages, fixed = TRUE)
+    cat(error_messages)
     stop("Compilation failed.", call. = FALSE)
   }
 
-
   shared_lib <- file.path(dir, "src", paste0(tools::file_path_sans_ext(basename(file)), .Platform$dynlib.ext))
   r_path <- file.path(dir, "R", "cpp11.R")
   brio::write_lines(r_functions, r_path)
```
jimhester, line 141: remove set_cpp_name

R/source.R
```diff
@@ -84,29 +84,24 @@ cpp_source <- function(file, code = NULL, env = parent.frame(), clean = TRUE, qu
...
 
-  # file not points to another location
-  file.copy(file, file.path(dir, "src", name))
-  #change variable name to reflect this
-  file <- file.path(dir, "src", name)
+  # file now points to another location
+  file.copy(file, file.path(new_path, name))
 
+  #change variable name to reflect this
+  file <- file.path(new_path, name)
```
jimhester, line 97: lets name this something else, new_file_path???

R/source.R
```diff
@@ -84,29 +84,24 @@ cpp_source <- function(file, code = NULL, env = parent.frame(), clean = TRUE, qu
...
-  if (base_cpp) {
-    name <- set_cpp_name(file)
-    package <- tools::file_path_sans_ext(name)
-  }
-  else {
-    # name and package might be different if cpp_source was called multiple times
-    name <- basename(file)
-    package <- tools::file_path_sans_ext(generate_cpp_name(file))
-  }
+  orig_path <- normalizePath(dirname(file), winslash = "/")
```
jimhester, line 90:
```suggestion
  orig_dir <- normalizePath(dirname(file), winslash = "/")
```

R/source.R
```diff
@@ -128,16 +123,19 @@ cpp_source <- function(file, code = NULL, env = parent.frame(), clean = TRUE, qu
...
 
   source_files <- normalizePath(c(file, cpp_path), winslash = "/")
-  res <- callr::rcmd("SHLIB", source_files, user_profile = TRUE, show = !quiet, wd = file.path(dir, "src"))
+  res <- callr::rcmd("SHLIB", source_files, user_profile = TRUE, show = !quiet, wd = new_path)
   if (res$status != 0) {
-    cat(res$stderr)
+    error_messages <- res$stderr
+
+    # Substitute temporary file path with original file path
+    error_messages <- gsub(file.path(new_path, basename(file)), file.path(orig_path, basename(file)), error_messages, fixed = TRUE)
```
jimhester, line 134:
```suggestion
    error_messages <- sub(file.path(new_path, basename(file)), file.path(orig_path, basename(file)), error_messages, fixed = TRUE)
```

jimhester:
```
filename <- basename(file)
new_path <- file.path(new_dir, filename)
orig_path <- file.path(orig_dir, filename)
error_messages <- sub(paste0(new_path, ": "), paste0(orig_path, ": "), error_messages, fixed = TRUE)
```

R/source.R
```diff
@@ -128,16 +123,19 @@ cpp_source <- function(file, code = NULL, env = parent.frame(), clean = TRUE, qu
...
+    cat(error_messages)
```
jimhester, line 135:
```suggestion
    cat(error_messages, file = stderr())
```

tests/testthat/test-source.R
```diff
@@ -197,4 +210,12 @@ test_that("check_valid_attributes returns an error if one or more registers is i
       }
 ')
   )
+
```
jimhester, line 213: :(

jimhester (commented):
After discussed changes feel free to merge!

jimhester (commented):
After discussed changes feel free to merge!

R/source.R
sbearrows: Check!

vspinu:
The consequence of this is that errors are printed now twice with different base paths. First ones with `xyz_N.cpp` and then the correct original files `xyz.cpp`. 

I understand that this is a hack to account for some editors (per #194) but there are editors out there which do understand tmp paths and redirect automatically to the correct file location. So those `_N.cpp` errors are quite a pain in the neck. 

Regardless of the editor I think it's cleaner and less confusing not to print `_N` in the first place. This was done in 4ef13b7e, which was factored out of the code base later in ad6a19ed994 and commits that followed. It's a bit hard for me to judge why that was done from the code history alone. So unless it's really critical for some reason to compile with those `_N` suffixes, could we please have that logic back?

vspinu:
While playing with it I think I know why the change was reverted. It's because the corner case of souring cpp11.cpp as per this [test](https://github.com/r-lib/cpp11/blob/d89706a83324a57c2ff25fb46bbe9cc3770caee0/tests/testthat/test-source.R#L39-L55). 

I think that corner case can just be addressed by changing only cpp11.cpp input file when detected. I am adding one more commit in #251 with tests and some clarification inline docs to avoid breaking this again.

PR #199, opened 2021-07-01
Feature remove redundant declarations

sbearrows:
Let me know what you think of my implementation!


R/register.R
```diff
@@ -258,12 +260,33 @@ get_call_entries <- function(path) {
...
     return("")
   }
-  res[seq(start, end)]
+
+  funs$package <- package
+
+  redundant <- as.character(glue_collapse_data(funs, 'extern SEXP _{package}_{name}', sep = '|'))
+
+  if (nchar(redundant) > 0) {
+    res <- res[!startsWith(res, redundant)]
```
jimhester, line 275: I'm not sure this will work properly if `redundant` had more than one search term, e.g.

```r
startsWith(c("foo", "bar"), "foo|bar")
#> [1] FALSE FALSE
# I think you need to use something like this
grepl("^(foo|bar)", c("foo", "bar"))
[1] TRUE TRUE
```

We should also have at least one test that tests for the case with multiple redundant names.

R/register.R
```diff
@@ -258,12 +260,33 @@ get_call_entries <- function(path) {
...
   if (length(start) == 0) {
     return("")
   }
-  res[seq(start, end)]
+
+  funs$package <- package
+
+  redundant <- as.character(glue_collapse_data(funs, 'extern SEXP _{package}_{name}', sep = '|'))
+
+  if (nchar(redundant) > 0) {
```
jimhester, line 274: I think `nzchar()` might be preferred here.

R/register.R
```diff
@@ -258,12 +260,33 @@ get_call_entries <- function(path) {
...
   end <- grep("};", res, fixed = TRUE)
 
   if (length(start) == 0) {
     return("")
   }
-  res[seq(start, end)]
+
+  funs$package <- package
+
+  redundant <- as.character(glue_collapse_data(funs, 'extern SEXP _{package}_{name}', sep = '|'))
```
jimhester, line 272: You shouldn't need `as.character()` here, glue returns a subclass of character so they should work everywhere without issue.

R/register.R
```diff
@@ -258,12 +260,33 @@ get_call_entries <- function(path) {
...
   start <- grep("/* .Call calls */", res, fixed = TRUE)
+
   end <- grep("};", res, fixed = TRUE)
 
   if (length(start) == 0) {
     return("")
   }
-  res[seq(start, end)]
+
+  funs$package <- package
```
jimhester, line 270: Do you need this? As far as I can tell you never use `funs$package` after this assignment.

tests/testthat/helper.R
```diff
@@ -12,6 +12,16 @@ pkg_path <- function(pkg) {
   dirname(pkg$.__enclos_env__$private$path)
 }
 
+get_funs <- function(path) {
+  all_decorations <- decor::cpp_decorations(path, is_attribute = TRUE)
+  get_registered_functions(all_decorations, "cpp11::register", quiet = FALSE)
+
```
jimhester, line 18: Extra whitespace

R/register.R
```diff
@@ -258,12 +260,33 @@ get_call_entries <- function(path) {
...
+
+  redundant <- as.character(glue_collapse_data(funs, 'extern SEXP _{package}_{name}', sep = '|'))
+
+  if (nchar(redundant) > 0) {
+    res <- res[!startsWith(res, redundant)]
+  }
+
+  mid <- grep("static const R_CallMethodDef CallEntries[] = {", res, fixed = TRUE)
+
+
```
jimhester, line 280: Extra whitespace

R/register.R
```diff
@@ -244,9 +245,10 @@ wrap_call <- function(name, return_type, args) {
   }
 }
 
-get_call_entries <- function(path) {
+get_call_entries <- function(path, funs, package) {
   con <- textConnection("res", local = TRUE, open = "w")
 
+
```
jimhester, line 251: Extra whitespace

R/register.R
```diff
@@ -70,7 +70,8 @@ cpp_register <- function(path = ".", quiet = FALSE) {
     cli::cli_alert_success("generated file {.file {basename(r_path)}}")
   }
 
-  call_entries <- get_call_entries(path)
+
```
jimhester, line 73: Extra whitespace

jimhester (approved):
overall looks good! I apologize in advance for my whitespace pedantry :)

After you have addressed the comments and added a NEWS entry feel free to (squash) merge this.

R/register.R
sbearrows: Ugh you're right thank you!

R/register.R
sbearrows: I used it in the next line to get the `extern SEXP _{package}_{name}` from a single tibble but I ended up sending funs$name directly and used `glue::glue()` for variables instead of `glue::glue_data`

PR #201, opened 2021-07-08
Feature converter ints doubles

### LINKED ISSUE #46: Permissive handling of integers in cpp11::integers
At the moment, coersion from an R numeric scalar to an `int` works nicely - doubles are converted to integers if they are representable as integers without truncation of the decimal part. However, the same is not true of a vector of numeric values into `cpp11::integers`:

```r
cpp11::cpp_source(
  code = c(
    "#include <cpp11.hpp>",
    "[[cpp11::register]]",
    "void test(int x, cpp11::integers y) {",
    "}"))
test(1L, 1L) # works, as expected
test(1, 1L)  # works, due to the double -> int conversion
test(1L, 1)  # fails with error "Invalid input type, expected 'integer' actual 'double'"
```

The same problem impacts passing a vector of integers to a function expecting doubles:

```r
cpp11::cpp_source(
  code = c(
    "#include <cpp11.hpp>",
    "[[cpp11::register]]",
    "void test2(cpp11::doubles y) {",
    "}"))
test2(seq_len(4))
#> Error in test2(seq_len(4)) : 
#>   Invalid input type, expected 'double' actual 'integer'
```

It would be nice if it is possible to relax this as practically this requires additional wrappers to the be added in either the C++ or the R code, especially as the name of the problematic variable can't be easily found from the error message.

sbearrows:
Might better to name them `as_doubles()` and `as_integers()` to differentiate from scalar `double` and `int`?? When I first was making it that sounded odd to me but now I'm not so convinced.

fixes #46


jimhester:
I think you are right, `as_doubles()` and `as_integers()` seems like a good idea to me.

jimhester:
Looks like you have a blank newline that is causing the format_check job to fail https://github.com/r-lib/cpp11/pull/201/checks?check_run_id=3024080555#step:5:11.

FWIW in RStudio you can enable showing whitespace characters with

Tools -> Global Options -> Code -> Display, check "Show whitespace characters".

Which could help spot these more easily.

sbearrows:
Okay I'm going to make those changes to function names. Any insights you might have on the ubuntu 3.4 error? I can't really tell where or why it's occurring.

jimhester:
I was able to reproduce it locally with R 3.4, it seems like an issue related to `cpp11::stop()`. That function requires unwind_protect, which IIRC has some issues on R versions before 3.5.

Switching the errors to throw regular C++ exceptions instead fixes the test failures.

I think it might be worth changing all the uses of `cpp11::stop()` in the cpp11 code to throw exceptions instead. In general I think `cpp11:stop()` seems to work fine for user code, but perhaps it would be better for the internal cpp11 code not to rely on it.

inst/include/cpp11/doubles.hpp
```diff
@@ -148,8 +148,7 @@ inline doubles as_double(sexp x) {
     return ret;
   }
 
-  // else
-  stop("Expected a numeric vector for");
+  throw type_error(INTSXP, TYPEOF(x));
```
jimhester, line 151: cpp11 defines a `type_error` class to generate a more informative error message for type issues, so we can use that here.

https://github.com/r-lib/cpp11/blob/23ae34fd6be84a3602e154fc3bf66897e117a38b/inst/include/cpp11/r_vector.hpp#L26-L39

sbearrows: Okay awesome thank you!

sbearrows:
Okay I will make a new issue to change instances of `cpp11::stop` to C++ style exceptions.

jimhester (approved):
LGTM, feel free to merge it once you are ready.

PR #204, opened 2021-07-13
Change internal `cpp11::stop()` to C++ exceptions

### LINKED ISSUE #203: Change instances of `cpp11::stop()` to C++ style exceptions
Using `cpp11::stop()` internally can cause issues on R versions before 3.5.

Consider switching all instances of `cpp11::stop()` to C++ exceptions instead.

sbearrows:
Fixes #203

First attempt so let me know if this looks correct.


jimhester:
These should throw some sort of classed exception object, e.g. one of https://en.cppreference.com/w/cpp/error/exception or a custom cpp11 exception class that derives from `std::exception`.

Possibly https://en.cppreference.com/w/cpp/error/invalid_argument would work for most of them, but you should use your judgement as to what is most appropriate for each case.

sbearrows:
Okay thank you!

jimhester (approved):
LGTM now, you can merge this whenever you are done

PR #207, opened 2021-07-19
Support more efficient nesting of unwind_protect

### LINKED ISSUE #141: There should be a way to opt out of unwind_protect for a block of code.
e.g. if you have a loop and want to put the protection outside around it rather than on each call.

Perhaps `unwind_protect()` should set a global flag, then we know that we don't need to unwind protect again?

jimhester:
The internal protection can be skipped if we are already in an
unwind_protect call.

Fixes #141

### DIFF
--- a/inst/include/cpp11/protect.hpp
+++ b/inst/include/cpp11/protect.hpp
@@ -183,8 +183,12 @@ static struct {
     static int* should_unwind_protect = nullptr;
 
     if (should_unwind_protect == nullptr) {
-      SEXP should_unwind_protect_sexp = Rf_allocVector(LGLSXP, 1);
-      R_PreserveObject(should_unwind_protect_sexp);
+      SEXP should_unwind_protect_sym = Rf_install("cpp11_should_unwind_protect");
+      SEXP should_unwind_protect_sexp = Rf_GetOption1(should_unwind_protect_sym);
+      if (should_unwind_protect_sexp == R_NilValue) {
+        should_unwind_protect_sexp = Rf_allocVector(LGLSXP, 1);
+        set_option(should_unwind_protect_sym, should_unwind_protect_sexp);
+      }
       should_unwind_protect = LOGICAL(should_unwind_protect_sexp);
       should_unwind_protect[0] = TRUE;
     }
@@ -207,6 +211,7 @@ preserved;
 template <typename Fun, typename = typename std::enable_if<std::is_same<
                             decltype(std::declval<Fun&&>()()), SEXP>::value>::type>
 SEXP unwind_protect(Fun&& code) {
+  REprintf("%s\n", *preserved.should_unwind_protect_ == TRUE ? "TRUE" : "FALSE");
   if (*preserved.should_unwind_protect_ == FALSE) {
     return std::forward<Fun>(code)();
   }
--- a/inst/include/cpp11/protect.hpp
+++ b/inst/include/cpp11/protect.hpp
@@ -26,194 +26,6 @@ class unwind_exception : public std::exception {
   unwind_exception(SEXP token_) : token(token_) {}
 };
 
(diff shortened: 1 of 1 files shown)


inst/include/cpp11/protect.hpp
```diff
@@ -26,6 +26,179 @@ class unwind_exception : public std::exception {
...
+      preserve_list = get_preserve_xptr_addr();
+      if (TYPEOF(preserve_list) != LISTSXP) {
+        preserve_list = Rf_cons(R_NilValue, R_NilValue);
+        R_PreserveObject(preserve_list);
+        set_preserve_xptr(preserve_list);
+      }
+    }
+
+    return preserve_list;
+  }
```
jimhester, line 180: All of this code was just moved here from lower down in the file, because it needs to be defined before the `unwind_protect()` calls below

inst/include/cpp11/protect.hpp
```diff
@@ -26,6 +26,179 @@ class unwind_exception : public std::exception {
...
+
+  static int* get_should_unwind_protect() {
```
jimhester, line 182: Here we define a global logical sexp to track if we are inside an unwind protect call.

We need to use a singleton managed by R rather than a global in the header file so we don't have multiple global variables in each compilation unit.

jimhester: It is possible we should move this code outside of this class, I put it here for convenience, since we were already using this same strategy for the preserve list.

inst/include/cpp11/protect.hpp
```diff
@@ -42,6 +221,7 @@ SEXP unwind_protect(Fun&& code) {
 
   std::jmp_buf jmpbuf;
   if (setjmp(jmpbuf)) {
+    *preserved.should_unwind_protect_ = TRUE;
```
jimhester, line 224: This one is important, otherwise we don't correctly set the flag after there is an error.

inst/include/cpp11/protect.hpp
```diff
@@ -26,6 +26,183 @@ class unwind_exception : public std::exception {
...
+      should_unwind_protect[0] = TRUE;
+    }
+
+    return &should_unwind_protect[0];
+  }
+
+  SEXP list_ = get_preserve_list();
+
+ public:
+  int* should_unwind_protect_ = get_should_unwind_protect();
```
jimhester, line 202: R_Boolean
```suggestion
  R_Boolean* should_unwind_protect_ = get_should_unwind_protect();
```

inst/include/cpp11/protect.hpp
```diff
@@ -66,6 +251,8 @@ SEXP unwind_protect(Fun&& code) {
   // unset it here before returning the value ourselves.
   SETCAR(token, R_NilValue);
 
+  *preserved.should_unwind_protect_ = TRUE;
+
```
jimhester, line 255: 😢

inst/include/cpp11/protect.hpp
```diff
@@ -34,6 +211,13 @@ class unwind_exception : public std::exception {
 template <typename Fun, typename = typename std::enable_if<std::is_same<
                             decltype(std::declval<Fun&&>()()), SEXP>::value>::type>
 SEXP unwind_protect(Fun&& code) {
+  REprintf("%s\n", *preserved.should_unwind_protect_ == TRUE ? "TRUE" : "FALSE");
```
jimhester, line 214: WTF

jimhester:
@sbearrows could you look this over, it should be a bit simpler than what we reviewed together yesterday.

I decided not to write an additional test for this, as it doesn't seem obvious to me how you would test it.

sbearrows (commented):
LGTM!

jimhester:
Thanks for reviewing!

PR #209, opened 2021-07-20
interface around {fmt} for `cpp11::stop()` and cpp11::warning()` message formatting

### LINKED ISSUE #169: Provide an interface around {fmt} for easier messaging
Basically wrappers around `RPrintf()` and `REPrintf()` as well as possibly `cpp11::stop()` and `cpp11::warning()`.

The named argument syntax is very similar to glue, and perhaps we could reuse our existing literal syntax in cpp11 for it.

https://fmt.dev/latest/index.html

sbearrows:
@jimhester Let me know what you think! Tests are pretty simple but we can add more if issues arise. 

Fixes #169


jimhester:
So I was thinking about this some this morning, and I think we will either have to use new function names (something like `stop_fmt()` for the fmt powered functions, or put them behind a `#ifdef CPP11_USE_FMT` flag that you have to opt into to use them.

This is because existing packages that are using `cpp11::stop()` are going to be using the sprintf style `%s` etc. message strings, and these won't work with fmt.

Another argument in favor of the `#ifdef` approach is that compiling the fmt code will incur some overhead, which some people may want to avoid.

Which way do you think would be best?

sbearrows:
I'd be in favor of the later! I think decreasing overhead is probably important to a lot of people using cpp11. What about you?

jimhester:
yeah that is the way I was leaning, though I worry if it is only opt-in then people may not realize it is an option / never use it.

sbearrows:
We could try to make a ruckus about? Adding examples in documentation?

Also, could we make fmt for `cpp11::stop()` and `cpp11::warning()` opt-in but make fmt for the new `cpp11::messages()` opt-out since it isn't present in previous versions? This might give people some breadcrumbs for the others? Not sure if that would be too confusing though.

cpp11test/src/test-protect.cpp
```diff
@@ -39,7 +39,7 @@ context("unwind_protect-C++") {
 
   test_that("stop throws an unwind_exception") {
     expect_error_as(cpp11::stop("error"), cpp11::unwind_exception);
-    expect_error_as(cpp11::stop("error: %s", "message"), cpp11::unwind_exception);
+    expect_error_as(cpp11::stop("error"), cpp11::unwind_exception);
```
bkietz, line 42:
```suggestion
    expect_error_as(cpp11::stop("error {}", "message"), cpp11::unwind_exception);
    expect_error_as(cpp11::stop("error {a}", fmt::arg("a", "message")), cpp11::unwind_exception);
```

inst/include/cpp11/protect.hpp
```diff
@@ -195,23 +198,27 @@ constexpr struct protect safe = {};
 inline void check_user_interrupt() { safe[R_CheckUserInterrupt](); }
 
 template <typename... Args>
-void stop [[noreturn]] (const char* fmt, Args... args) {
-  safe.noreturn(Rf_errorcall)(R_NilValue, fmt, args...);
+void stop [[noreturn]] (const char* fmt_arg, Args... args) {
```
bkietz, line 201: Since {fmt} can accept (for example) a std::string without copying it, it'd be better to change these parameter packs to const references
```suggestion
void stop [[noreturn]] (const char* fmt_arg, const Args&... args) {
```

inst/include/cpp11/protect.hpp
```diff
@@ -195,23 +198,27 @@ constexpr struct protect safe = {};
...
+  std::string msg = fmt::format(fmt_arg, args...);
+  safe.noreturn(Rf_errorcall)(R_NilValue, "%s", msg.c_str());
 }
 
 template <typename... Args>
-void stop [[noreturn]] (const std::string& fmt, Args... args) {
-  safe.noreturn(Rf_errorcall)(R_NilValue, fmt.c_str(), args...);
+void stop [[noreturn]] (const std::string& fmt_arg, Args... args) {
+  std::string msg = fmt::format(fmt_arg, args...);
```
bkietz, line 208: Also, it's unfortunate that we're losing the `%x` format specifier for `SEXP`s. This could be recoverable by introducing a helper function:
```c++
// pass through by default
template <typename T>
const T& as_formattable(const T& value) { return value; }

// specialization for SEXP: 
std::string as_formattable(SEXP sexp) { return gettextf("%x", sexp); }
```

bkietz (changes requested):

inst/include/cpp11/protect.hpp
jimhester: I think you would use `cpp11::stop("{0:x}", my_sexp)` if you want to print the hexadecimal address for a sexp.

bkietz: Oh, I misunderstood the purpose of `%x`. I thought it gave a string representation of the pointed value rather than a representation of the pointer itself. Let me rephrase: would it be desirable to intercept `SEXP` pointers and replace them with `capture.output(print(value))` or so? It'd probably be better to include that as a `sexp::to_string()` so that we can write

```c++
cpp11::stop("Invalid: {}", sexp(value).to_string());
```

jimhester: Ah, maybe, though it is not always clear what output would be preferred for a given SEXP, possibly something like that returned from `R_inspect()` might be the most useful for any SEXP type, though the default arguments can generate a lot of data for deeply nested objects.

```r
> .Internal(inspect(1:10))
@7ff6163c1388 13 INTSXP g0c0 [REF(65535)]  1 : 10 (compact)
```

inst/include/cpp11/protect.hpp
```diff
@@ -195,23 +198,27 @@ constexpr struct protect safe = {};
 inline void check_user_interrupt() { safe[R_CheckUserInterrupt](); }
 
 template <typename... Args>
-void stop [[noreturn]] (const char* fmt, Args... args) {
-  safe.noreturn(Rf_errorcall)(R_NilValue, fmt, args...);
+void stop [[noreturn]] (const char* fmt_arg, Args... args) {
+  std::string msg = fmt::format(fmt_arg, args...);
```
bkietz, line 202: Wasn't aware of [this restriction](https://github.com/r-lib/cpp11/pull/209/checks?check_run_id=3128722782#step:12:76), adding perfect forwarding here should clean it up:
```suggestion
void stop [[noreturn]] (const char* fmt_arg, Args&&... args) {
  std::string msg = fmt::format(fmt_arg, std::forward<Args>(args)...);
```

bkietz (changes requested):

jimhester (approved):

jimhester:
LGTM, I think this can be merged once the tests are green.

PR #214, opened 2021-07-26
Fix Windows error

sbearrows:
Attempt to fix windows error


jimhester:
Maybe try creating a branch and PR with `usethis::pr_init()` rather than via the git command line.

I suspect the reason this is failing on windows has to do with initializing this in global scope https://github.com/r-lib/cpp11/blob/ac7374e81bb9e8fe35b21424c53e21b91b00c9e4/inst/include/cpp11/function.hpp#L78.

Maybe try putting it as a static variable inside the functions instead.

jimhester:
Closed by #215

PR #223, opened 2021-08-11
Don't rely on global constructors for the preserved list

### LINKED ISSUE #224: cpp11::preserved relies on ordered global construction
`cpp11::preserved` relies on the `list_` data member having a well-defined value at the time `preserved` is accessed. Since the order of construction of globals is not guaranteed in c++, that can lead to undefined behavior if you're constructing (for example) a global `cpp11::strings` since that appends to `preserved` in the constructor.

```c++
namespace {

// the value of `cpp11::preserved.list_` is not guaranteed here!
cpp11::strings kColorStrings = cpp11::writable::strings({
  "red",
  "green",
  "blue",
});

} // namespace
```

bkietz:
fixes #224

### DIFF
--- a/inst/include/cpp11/protect.hpp
+++ b/inst/include/cpp11/protect.hpp
@@ -313,9 +313,7 @@ static struct {
 
     PROTECT(obj);
 
-    if (TYPEOF(list_) != LISTSXP) {
-      list_ = get_preserve_list();
-    }
+    SEXP list_ = get_preserve_list();
 
     // Add a new cell that points to the previous end.
     SEXP cell = PROTECT(Rf_cons(list_, CDR(list_)));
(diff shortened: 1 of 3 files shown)


bkietz:
See also: https://github.com/apache/arrow/pull/10908

inst/include/cpp11/protect.hpp
```diff
@@ -313,9 +313,7 @@ static struct {
 
     PROTECT(obj);
 
-    if (TYPEOF(list_) != LISTSXP) {
```
jimhester, line 316: The only worry I have about this is the increased overhead to lookup the preserve list on each call to `insert()`

Wonder if we should make this static so it is only initialized once in each compilation unit?

jimhester (approved):

inst/include/cpp11/protect.hpp
bkietz: Alright, I can do that

PR #229, opened 2021-08-11
Columns support for matrix class

alyst:
This draft PR adds `matrix::column` class in full analogy to `matrix::row` class and updates the matrix API so that both rows and columns could be accessed. I just added it because my use-case required column-wise access.
I'm not sure about the proposed matrix API, it was just the easiest to implement.
Alternatively one could think about `matrix::row_vector` proxy class and `matrix::rows()` method that would provide row access (begin/end/[] etc) and `matrix::column_vector` that would provide symmetric columns access.

I also added the required types to `matrix::row::iterator` so that it actually conforms to the iterator concept.

### DIFF
--- a/inst/include/cpp11/matrix.hpp
+++ b/inst/include/cpp11/matrix.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <iterator>
 #include <string>  // for string
 
 #include "cpp11/R.hpp"         // for SEXP, SEXPREC, R_xlen_t, INT...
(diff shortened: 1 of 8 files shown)


alyst:
The tests are failing because the matrix::row-related API has changed. I will fix the tests once there is a decision how this API should actually look like.

jimhester:
An alternative interface would be to let the user to supply if they want to use column or row based indexing as part of the matrix constructor, either as a template parameter or an argument.

I think generally users only want to iterator over rows or columns, and very rarely do both with the same matrix. If they do want to use both in the same application they could create two matrix objects pointing at the same underlying data.

alyst:
@jimhester I've tried to implement your suggestion.

Now the `matrix` class has the 3rd template parameter, `S` (how the matrix is sliced), which could be either `by_row` or `by_column`.
So all fully specialized matrix types are duplicated: there is `doubles_row_matrix` and `doubles_column_matrix`.
The `matrix::row` class is replaced by `matrix::slice`, which uses `matrix::slice_xxx()` methods to define how it accesses the matrix elements.
`matrix<V,T,S>` is inherited from `matrix_slice<S>`, where the `slice_xxx` methods are actually defined (there are `matrix_slice<by_row>` and `matrix_slice<by_column>` specializations).

It's been a while since I wrote something in C++, I hope I have not overcomplicated these specialization things.

If this approach looks good, I can start updating the tests.

alyst:
@jimhester the tests pass now

inst/include/cpp11/matrix.hpp
```diff
@@ -86,24 +191,32 @@ class matrix {
...
-  row end() { return {*this, nrow_}; }
+  slice_iterator begin() { return {*this, 0}; }
+  slice_iterator end() { return {*this, nslices()}; }
 };
 
-using doubles_matrix = matrix<r_vector<double>, double>;
-using integers_matrix = matrix<r_vector<int>, int>;
-using logicals_matrix = matrix<r_vector<r_bool>, r_bool>;
-using strings_matrix = matrix<r_vector<r_string>, r_string>;
+template <typename S = by_column>
```
jimhester, line 202: I think the default access pattern would need to be by row, as this is what cpp11 currently supplies.

alyst: I can change that. I just have thought that cpp11 is in active development and have not reached 1.0.0 yet, so changes like this are still possible. The reason is - column slices have continuous memory layout, so it would be nice to select more efficient alternative by default. And the fact that there were some issues with matrix iterators suggests that so far there are not so many matrix users whose code might be broken.

jimhester: That is true, let me ask around and see what people might prefer.

jimhester: Alright since R generally does column major ordering by default I think having cpp11 match that makes the most sense. So I think leaving it as you have it now is OK

alyst: Great! Thanks for double-checking.
I've applied some cosmetics and squashed the commits to clean-up the log.

jimhester:
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Otherwise I think this looks great, thanks for working on it!

alyst:
@jimhester done

jimhester:
Thanks again for working on this!

Thanks a bunch!

PR #230, opened 2021-08-11
cpp_register(): use_package by default

alyst:
For some reason the current symbol-based approach didn't work for me (R 4.0.5) resulting in `Symbol not found`  errors for `.Call()`.

There is already support for an alternative `.Call(..., PACKAGE='...')` approach, I just extended it to `cpp_register()` function and made it the default.


jimhester:
Thanks for working on this, but I am confident that the symbols will work regardless of the R version and we use symbols for packages for a reason.

I suspect there is something else going wrong with registration in your package, would it be possible for you to provide a link to it?

alyst:
@jimhester It could be because I've missed `.registration = TRUE` in my `useDynLib()` directive. Now the symbol-based wrappers seem to work.

For this PR I can set the `cpp_register(..., use_package = FALSE, ...)`, so that the *cpp11* behavior would not change,
but still there would be a more complete support for generating the wrappers that explicitly specify the package.

jimhester:
Sure we can expose the argument, though in general I don't think most packages would ever need to use it.

Also FWIW if you use `usethis::use_cpp11()` to setup cpp11 for your package it will add a roxygen annotation to include `.registration = TRUE` in your namespace directive.

alyst:
Ok, I've changed to `use_package=FALSE`.

DavisVaughan:
I've decided not to do this for now, as symbols work everywhere and are the most useful thing to use

PR #231, opened 2021-08-12
Fix semantics of + and += for r_vector::iterator

alyst:
IIRC, + should return new object, and += should modify the existing one.


jimhester:
Thanks, can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

alyst:
@jimhester I've rebased and addedd the NEWS entry.

jimhester:
Thanks a ton!

Issue #246, opened 2021-10-27
How to set up `CPPFLAGS` with `cpp11::cpp_source`


vspinu:
Sorry if obvious, but I cannot figure out a way to provide compiler flags to `cpp_soruce` besides modifying the `~/.R/Makevars`. A local `Makevars` file is not picked up.

paleolimbot:
You can also temporarily set the `PKG_CPPFLAGS` environment variable (totally unverified hack on my end). For example, I use this to prototype C++ linking to my local development version of Arrow:

``` r
withr::with_envvar(
  list(
    PKG_CXXFLAGS = paste0("-I", Sys.getenv("ARROW_HOME"), "/include"),
    PKG_LIBS = paste0("-L", Sys.getenv("ARROW_HOME"), "/lib", " -larrow")
  ),
  cpp11::cpp_source("2021_11_05_copy_files.cpp")
)
```

PR #251, opened 2021-11-23
Cpp source include fixes

### LINKED ISSUE #216: How to include custom headers on top of generated cpp11.cpp?
My function returns `external_pointer<MyClass>`. How can I get a definition of `MyClass` into cpp11.cpp? 

I was hoping for a declaration like `[cpp11:include]` which I can put in front of some statement and it will be automatically included in `cpp11.cpp`. Is there such a thing?

vspinu:
Two improvements to `srouce_cpp()`:

  - Honor `<basename>_types.h` or `<basename>_types.hpp` (I included both extensions because editor support might be different for different extensions. And I feel it's cleaner to use hpp for cpp code).  Fixes #216 
  - Add file's directory to the include path. (addresses most common use case of #246 without dealing with CPPFLAGS trickery)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # cpp11 (development version)
 
 * Type files `<basename>_type.h` is no honored by `source_cpp()` (#216). 
+* The directory of the sourced file is included into the include path during `source_cpp()` compilation. 
 
 # cpp11 0.4.1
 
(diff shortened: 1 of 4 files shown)


vspinu:
Re-add cleaning of `_N.cpp` suffixing logic  originally done in #176 (4ef13b7) as per [comment](https://github.com/r-lib/cpp11/pull/198#issuecomment-981135699).

PR #269, opened 2022-03-16
Basic C++ to C++ API generation

nx10:
This PR adds functionality to generate C/C++ APIs which can be accessed from other packages.

## Motivation

- Fast communication between packages
- Manually creating the required registration and headers is cumbersome, error-prone and violates DRY principles

## Implementation

Functions annotated with `[[cpp::external]]` are registered in `src/cpp11.cpp` and a header `inst/include/{package}_api.h` is created which can be included in other packages after `LinkingTo:` and importing the package.

The generated code is inspired by [similar APIs in `systemfonts`](https://github.com/r-lib/systemfonts/blob/main/inst/include/systemfonts.h) and some other packages.

Features:

- There are two optional parameters for renaming the exported function and adding a doc comment `[[cpp::external("other_name", "Some documentation comment")]]`
- `inst/include/{package}_types.h` will be included in the header if it exists.

## ToDo

- Unit testing

## Discussion

Please let me know if this is in scope for `cpp11` (I hope so), or if you have any comments or suggestions. 

Also thank you for your work on the `cpp11` package!


jimhester:
cc @thomasp85 who might have interest in this approach.

PR #272, opened 2022-05-10
Allow passing number of rows from outside

krlmlr:
The internal calc_nrow() doesn't (want to) implement the full vctrs logic, this offers a way for callers to construct a valid data frame in all corner cases.


krlmlr:
The test failures now seem consistent with those seen on the main branch.

krlmlr (commented):
Thanks for taking this on, great to see this implemented!

PR #279, opened 2022-08-22
Update GitHub Actions workflows, and make necessary tweaks to support them

DavisVaughan:
This PR was originally intended to update the GitHub Actions workflow files in cpp11, but doing so proved to be quite challenging since they haven't been updated in awhile.

Note that there are more files changed here than just the workflow files, because we needed to tweak some of the actual cpp11 tests to accommodate some other problems. I will point them out with inline comments describing them in detail.

You should also be able to look at the individual commits to track the changes, I have tried to make them meaningful.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -29,14 +29,14 @@ jobs:
           - {os: macOS-latest,   r: 'release'}
           - {os: windows-latest, r: 'release'}
           - {os: windows-latest, r: '3.6', rspm: "https://packagemanager.rstudio.com/cran/latest"}
-          - {os: ubuntu-18.04,   r: 'devel', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.1.0 (ubuntu-18.04) R (4.1.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
-          - {os: ubuntu-18.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", custom: 'no-cpp11test'}
-          - {os: ubuntu-18.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", custom: 'gcc 4.8'}
-          - {os: ubuntu-18.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: '3.6',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: '3.5',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: '3.4',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          - {os: ubuntu-20.04,   r: 'devel', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.1.0 (ubuntu-20.04) R (4.1.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
+          - {os: ubuntu-20.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", custom: 'no-cpp11test'}
+          - {os: ubuntu-20.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", custom: 'gcc 4.8'}
+          - {os: ubuntu-20.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          - {os: ubuntu-20.04,   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          - {os: ubuntu-20.04,   r: '3.6',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          - {os: ubuntu-20.04,   r: '3.5',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          - {os: ubuntu-20.04,   r: '3.4',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
 
     env:
       RSPM: ${{ matrix.config.rspm }}
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -1,18 +1,14 @@
-# NOTE: This workflow is overkill for most R packages
-# check-standard.yaml is likely a better choice
-# usethis::use_github_action("check-standard") will install it.
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 #
-# For help debugging build failures open an issue on the RStudio community with the 'github-actions' tag.
-# https://community.rstudio.com/new-topic?category=Package%20development&tags=github-actions
+# NOTE: This workflow is overkill for most R packages and
+# check-standard.yaml is likely a better choice.
+# usethis::use_github_action("check-standard") will install it.
 on:
   push:
-    branches:
-      - main
-      - master
+    branches: [main, master]
   pull_request:
-    branches:
-      - main
-      - master
+    branches: [main, master]
 
 name: R-CMD-check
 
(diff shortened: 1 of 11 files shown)


.github/workflows/R-CMD-check.yaml
```diff
@@ -27,53 +23,35 @@ jobs:
...
-          - {os: ubuntu-18.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: '3.6',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: '3.5',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-          - {os: ubuntu-18.04,   r: '3.4',     rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
+          # Use 3.6 to trigger usage of RTools35
+          - {os: windows-latest, r: '3.6'}
+
+          # Use older ubuntu to maximise backward compatibility
+          - {os: ubuntu-20.04,   r: 'devel', http-user-agent: 'release'}
```
IndrajeetPatil, line 32: IINM, the new recommendation is to use the latest Ubuntu, i.e. `os: ubuntu-latest`, instead of `os: ubuntu-20.04`.

cf. https://github.com/r-lib/actions/issues/606

DavisVaughan: cpp11 is probably going to be one of the projects that still tries somewhat hard to support older ubuntu versions

DavisVaughan:
Blocked by a few more things:

- We need dev testthat to avoid the warnings generated by this https://github.com/r-lib/testthat/issues/1672, but installing dev testthat is failing in pak on the Windows build we care about https://github.com/r-lib/pak/issues/407
- We have to make a decision about whether or not we want to continue supporting gcc 4.8 in light of the fact that Ubuntu 20.04 no longer supports it out of the box, and we don't have access to an easy Ubuntu 18.04 worker anymore

.github/workflows/R-CMD-check.yaml
```diff
@@ -27,92 +23,74 @@ jobs:
       matrix:
         config:
           - {os: macOS-latest,   r: 'release'}
+
           - {os: windows-latest, r: 'release'}
-          - {os: windows-latest, r: '3.6', rspm: "https://packagemanager.rstudio.com/cran/latest"}
-          - {os: ubuntu-18.04,   r: 'devel', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.1.0 (ubuntu-18.04) R (4.1.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
```
DavisVaughan, line 32: GitHub Actions is no longer supporting Ubuntu 18.04, so we switched to 20.04.

Note that r-lib/actions actually just switched to `ubuntu-latest` as the default, but I think for cpp11 it is worth trying to continue to test on "older" Ubuntu versions like 20.04 https://github.com/r-lib/actions/pull/612

.github/workflows/R-CMD-check.yaml
```diff
@@ -27,92 +23,74 @@ jobs:
...
+        # https://github.com/r-lib/cpp11/pull/78
+        # https://www.rstudio.com/about/platform-support/
+        # Ubuntu 20.04 technically dropped support for gcc 4.8, so we have to
+        # add old archives back in manually to install it
+        # https://github.com/r-lib/cpp11/pull/279
         if: matrix.config.custom == 'gcc 4.8'
         run: |
+          echo "deb http://dk.archive.ubuntu.com/ubuntu/ xenial main" | sudo tee -a /etc/apt/sources.list
+          echo "deb http://dk.archive.ubuntu.com/ubuntu/ xenial universe" | sudo tee -a /etc/apt/sources.list
+          sudo apt update
```
DavisVaughan, line 69: I have retained support for testing gcc 4.8, but it wasn't easy. The move to Ubuntu 20.04 means that `sudo apt-get install -y g++-4.8` doesn't work out of the box anymore. 

The way to fix this is to add older archives into the `sources.list` manually. This was proposed on a number of stack overflow questions, like these
https://stackoverflow.com/questions/68805002/i-need-to-install-gcc4-9-on-ubuntu-20-04-matlab-mex
https://askubuntu.com/questions/1036108/install-gcc-4-9-at-ubuntu-18-04

Corresponds to this commit https://github.com/r-lib/cpp11/pull/279/commits/6a2994ad966ae1372b815a4be19af7f80e9ce692

.github/workflows/R-CMD-check.yaml
```diff
@@ -27,92 +23,74 @@ jobs:
...
           sudo apt-get install -y g++-4.8
           mkdir ~/.R/
-          echo $'CXX1X=g++-4.8\nCXX11=g++-4.8' >> ~/.R/Makevars
+          echo $'CXX1X=g++-4.8\nCXX11=g++-4.8\nCC=gcc-4.8' >> ~/.R/Makevars
```
DavisVaughan, line 72: The addition of `CC` is fairly tricky. If you don't add this, the FAQ vignette fails to build with errors about `-fdiagnostics-color=always` being a flag that isn't supported. Note that the FAQ vignette has `{cpp11}` knitr chunks that it compiles, and while it was compiling those chunks it was indeed setting the `-fdiagnostics-color=always` flag.

Some research turned up that this flag was only supported in gcc >=4.9, so gcc is right that it shouldn't be there.

It turns out this flag was being added by pkgbuild during `pkgbuild::build()`.
https://github.com/r-lib/pkgbuild/blob/80762db02d9273997ed6234f654633359269cc9e/R/compiler-flags.R#L51-L58
https://github.com/r-lib/pkgbuild/blob/80762db02d9273997ed6234f654633359269cc9e/R/build.R#L54

It gets added if it looks like the compiler supports colored diagnostics, i.e. the `has_compiler_colored_diagnostics()` check there. This helper works by attempting to compile a simple C program with `-fdiagnostics-color=always` set in the Makevars https://github.com/r-lib/pkgbuild/blob/80762db02d9273997ed6234f654633359269cc9e/R/compiler-flags.R#L76. If the compilation fails, we assume there is no color support. The problem is that it attempts this on a _C_ program which uses a compiler that does support color, but the _C++_ compiler we set through `CXX11` does not.

So my fix for this is to just force the C compiler to also be gcc 4.8 so this helper returns `FALSE`.

Corresponds to this commit https://github.com/r-lib/cpp11/pull/279/commits/f59c83342f4fc9898990e01c332011f16bf3b392

.github/workflows/pkgdown.yaml
```diff
@@ -1,35 +1,48 @@
...
+      - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: local::. local::cpp11test pkgdown
+          extra-packages: any::pkgdown, local::., local::cpp11test
           needs: website
 
-      - name: Deploy package
+      - name: Build site
         env:
           CPP11_EVAL: true
```
DavisVaughan, line 38: In particular, made sure to retain this!

.github/workflows/pkgdown.yaml
```diff
@@ -1,35 +1,48 @@
...
-      - uses: r-lib/actions/setup-r@v1
+      - uses: r-lib/actions/setup-r@v2
         with:
           use-public-rspm: true
 
-      - uses: r-lib/actions/setup-r-dependencies@v1
+      - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: local::. local::cpp11test pkgdown
+          extra-packages: any::pkgdown, local::., local::cpp11test
```
DavisVaughan, line 33: Retained installation of cpp11test here

.github/workflows/test-coverage.yaml
```diff
@@ -1,61 +1,38 @@
...
-      - name: Install dependencies
+      - name: Install cpp11test
         run: |
-          pak::local_install(dependencies = FALSE)
+          options(warn = 2)
           pak::local_install_dev_deps("cpp11test", dependencies = TRUE)
-          install.packages("cpp11test", repos = NULL, INSTALL_opts = "--install-tests")
-          pak::local_install(dependencies = TRUE)
-          pak::pkg_install("covr")
+          install.packages("cpp11test", repos = NULL, INSTALL_opts = "--install-tests", type = "source")
```
DavisVaughan, line 33: Mainly just syncing this to what is in `R-CMD-check.yaml`

DESCRIPTION
```diff
@@ -60,3 +60,5 @@ Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.2
 SystemRequirements: C++11
+Remotes:
+    r-lib/testthat
```
DavisVaughan, line 64: There was a testthat issue that was causing some warnings when compiling cpp11test on Windows with the "latest" version of R. Since we use `options(warn = 2)` before compiling it was causing the build to fail https://github.com/r-lib/testthat/issues/1672

Corresponds to this commit https://github.com/r-lib/cpp11/pull/279/commits/7a479d84317fed184e8ee24c74cfbeaa4c5b356c

cpp11test/src/test-doubles.cpp
```diff
@@ -184,7 +185,12 @@ context("doubles-C++") {
 
 #if defined(__APPLE__) && defined(R_VERSION) && R_VERSION >= R_Version(3, 5, 0)
   test_that("writable::doubles(ALTREP_SEXP)") {
-    SEXP x = PROTECT(Rf_coerceVector(R_compact_intrange(1, 5), REALSXP));
+    // ALTREP compact-seq
+    auto seq = cpp11::package("base")[":"];
+    cpp11::sexp range = seq(cpp11::as_sexp(1), cpp11::as_sexp(5));
+
+    SEXP x = PROTECT(Rf_coerceVector(range, REALSXP));
+    expect_true(ALTREP(x));
```
DavisVaughan, line 193: It turns out that `R_compact_intrange()` wasn't supposed to be exposed to us, and R core has moved it to `Defn.h` in R 4.2.0 (or maybe 4.2.1, not sure)
https://github.com/wch/r-source/commit/86b4a0641e889808ce204e28dd8bd19bbd1c8cbb

I've replaced it with a call to `1:5`, which does make an ALTREP object that we can use here, and seemed like a reasonable way to make one for now (with the alternative being that we create a full altrep dummy class in cpp11test to test with)

Corresponds to this commit https://github.com/r-lib/cpp11/pull/279/commits/29cc8339811fe1eb91f2d7a4886111d9f881dd33

cpp11test/src/test-protect.cpp
```diff
@@ -1,3 +1,16 @@
...
+#ifdef __clang__
+# pragma clang diagnostic push
+# pragma clang diagnostic ignored "-Wunused-variable"
+#endif
+
+#ifdef __GNUC__
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wunused-variable"
+#endif
+// clang-format on
```
DavisVaughan, line 12: Another thing that showed up on the Windows build with the "latest" version of R while compiling cpp11test was a warning about `cpp11::preserved` being an unused variable in this `test-protect.cpp` file.

I guess that makes sense, because it isn't used anywhere in this test file but is created as a `static struct` in the header file so each compilation unit gets a copy of it and is "expected" to use it. Not sure why we don't get this warning on other OSes, but I just copied over the same pragma from `protect.cpp` that is also used to ignore some unused variable warnings
https://github.com/r-lib/cpp11/blob/cb1a6884990adb6d5c241e86fcfd9293e122c5fc/inst/include/cpp11/protect.hpp#L301
https://github.com/r-lib/cpp11/blob/cb1a6884990adb6d5c241e86fcfd9293e122c5fc/cpp11test/src/protect.cpp#L37

Corresponds to this commit https://github.com/r-lib/cpp11/pull/279/commits/0fbdcb1ddb35146385689a89ade4bc061ede5f75

DavisVaughan:
@jimhester would you mind reviewing this PR for me? There are a number of things here where I want to be sure that I am correctly preserving what was already there, and I think having a review from you would be very helpful for that.

I would have broken it up into smaller individual PRs if I could, but it seemed more useful to have 1 PR with well defined commits that results in all builds passing again.

.github/workflows/R-CMD-check.yaml
jimhester: yeah, that sounds right, I wonder if pkgbuild should test both the CC and CXX points.

Alternatively perhaps it is better practice to use the same toolchain for C and C++ anyway (even though there is no C code being compiled here)

.github/workflows/R-CMD-check.yaml
jimhester: Calling out the fact that RStudio are doing this to support CentOS7, I wonder if it would make sense to move to using docker for CentOS 7 explicitly, in order to better replicate the environment you are trying to support, IIRC you would have to use Docker for everything then, so maybe not worth it, or you could run this in a separate workflow...

.github/workflows/R-CMD-check.yaml
DavisVaughan: > I wonder if pkgbuild should test both the CC and CXX points.

Gabor and I are talking about exactly that in https://github.com/r-lib/pkgbuild/issues/141

DESCRIPTION
jimhester: wonder if we should pin this to a commit SHA rather than having it float, and remove the pin after the changes are on CRAN.

I worry that having it float with the HEAD would cause builds to fail due to unrelated testthat development

cpp11test/src/test-doubles.cpp
jimhester: Makes sense, wonder if it is more readable to use `seq` rather than `:`, though the former is a S3 generic, so not quite the same thing, I don't think it will ever matter in practice.

jimhester (approved):
Looks good to me, sorry about all the finickiness, cpp11 is kind of in a weird place with its interaction with CI

.github/workflows/R-CMD-check.yaml
DavisVaughan: I opened https://github.com/r-lib/cpp11/issues/281#issue-1350885749 to consider this

Arrow uses a docker image in this way, so we could probably do the same

PR #286, opened 2022-09-19
Ext ptr tag

thk686:
Adding the ability to get/set tag on external pointer class.


romainfrancois:
I've messed with the history in this PR. I'll extract the relevant changes in `external-pojnter.hpp` and `test-external-pointer.cpp` and do it again. Sorry about the inconvenience.

thk686:
No worries. Actually, I ended up not using tags in my experiment. But it
would be helpful to wrap setting the protect object as I use that a lot.

romainfrancois:
Ok, let me close this then, and I'll pick up #287 in a new PR.

PR #310, opened 2023-03-14
Respond to upcoming changes in `glue::glue_collapse()`

jennybc:
In the next version of glue, `glue::glue_collapse()` will never return `character()` but instead will return `""` for empty inputs. For more detail see https://github.com/tidyverse/glue/pull/295.

This makes the necessary change inside cpp11 and places 2 temporary skips in the tests so cpp11 can be released before glue. Interestingly, this actually makes the tests match their existing descriptions.

I won't release glue for at least two weeks (so: March 31) and it will likely be later.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,6 @@ Suggests:
     decor,
     desc,
     ggplot2,
-    glue,
     knitr,
     lobstr,
     mockery,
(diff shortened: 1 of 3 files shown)


jennybc:
The CI failure doesn't appear to be associated with what's going on in this PR.

romainfrancois (approved):
Thanks @jennybc. I'll try to release `cpp11` before `glue`

PR #316, opened 2023-05-16
Update README.md with note on imports

pepijn-devries:
When using functions from other packages using the cpp11 `package` function, I noticed that when calling such functions R would get stuck and doesn't throw errors or warnings. It took me a while to realise that these functions weren't imported to the namespace. After importing the functions, everything worked smoothly. I've added this note to help others avoid running into the same problem.


pepijn-devries:
Thanks for the response. Yes an error-message will definitely help here.

PR #327, opened 2023-07-26
Remove `cpp11_should_unwind_protect` altogether

### LINKED ISSUE #325: `unwind_protect()` global option is falsely assumed to be immutable
Fixing https://github.com/r-lib/cpp11/pull/299, which ensures that we are always referencing the R global option `cpp11_should_unwind_protect` to determine if we need to unwind-protect or not, revealed a big issue with how this is currently set up.

CRAN emailed us about this, and actually reverted cpp11 0.4.4 back to 0.4.3 code (but bumping the version to 0.4.5) because this was such a problem.

An easy way to demonstrate the issue is to install cpp11 0.4.4 and run `devtools::check()` on the hdpGLM package (there are many others that fail with 0.4.4, including FIESTA MIMSunit amt arealDB campsis ggPMX hdpGLM mpathsenser tidyfit). You should get `recursive gc` issues and R will eventually crash.

---

The core problem is that cpp11 expects that the `cpp11_should_unwind_protect` global option _never_ changes. This is because we take a `static` pointer to its underlying `logical(1)` data, which we modify directly from C++ as needed.

This seems reasonable, but many package authors do this:

```r
opt <- options()
on.exit(options(opt))
```

which _does_ create a copy of every individual option value, including `cpp11_should_unwind_protect`, invalidating our pointer.

In particular it is done here for hdpGLM, which then calls `tidyr::gather()`, which uses C++ code written with cpp11, and that is always roughly where the crash occurs:
https://github.com/DiogoFerrari/hdpGLM/blob/4ba7b7c88a9a2bea344f13ae373f84b48a1d5a62/R/src_summaries.R#L1546-L1547

---

A possible fix for this specific issue that Kevin suggested is to instead use an environment in the global option, because those won't be copied, and then put our cpp11 specific values in that environment.

One thing we also discussed is whether changing this will affect other already built packages that linked against "old" cpp11. Those packages will be looking for and modifying `cpp11_should_unwind_protect`, while any packages built with "new" cpp11 will be looking for `cpp11_environment` (or whatever we call it). This isn't ideal, but _might_ be okay. At the absolute worst case the package would need to be reinstalled with "new" cpp11.

We could also consider versioning the slots inside the environment based on the current cpp11 version.

So it may look something like:

```r
cpp11_environment <- new.env()

cpp11_environment_0_4_6 <- new.env()
cpp11_environment_0_4_6[["should_unwind_protect"]] <- TRUE

cpp11_environment[["0.4.6"]] <- cpp11_environment_0_4_6

options(cpp11_environment = cpp11_environment)
```

It is also possible that we _don't_ want to do that, however. The following cpp11 code can crash if `should_unwind_protect` isn't set up correctly:

```cpp
[[cpp11::register]]
void test() {
  cpp11::unwind_protect([&] {
    cpp11::unwind_protect([&] {
      Rf_error("oh no!");
    });
  });
}
```

If `test()` is called from R:
- It goes through `.Call` and our wrapper, which sets up the `BEGIN_CPP11` and `END_CPP11` macros
- It calls `unwind_protect()` which calls `R_UnwindProtect()`, a C function!
- Inside `R_UnwindProtect()`, we call `unwind_protect()`
- From there we `Rf_error()`
- The inner `unwind_protect()` catches that C error and promotes it to a C++ exception which is `throw`n.
- That is thrown _across_ C stack frames, because we are inside the outer `R_UnwindProtect()` and no try/catch was set up within that to catch the unwind exception. The only try/catch is that most outer one set up by the macros
- That is UB and R crashes due to an uncaught exception

It is technically possible to be in a scenario where the outer `unwind_protect()` comes from a package built with one version of cpp11 and the inner `unwind_protect()` comes from a package built with another version of cpp11. If both calls think they should unwind protect (because they are looking at different versions of `should_unwind_protect`), then we crash.

### LINKED ISSUE #326: Nested `unwind_protect()` calls going through R can be very dangerous
Assume you have cpp11 0.4.4 installed, with a "working" version of the `cpp11_should_unwind_protect` R global option ("working" in the sense that if you enter a nested unwind-protect call, then it notices that there is an outer unwind-protect present and decides not to unwind-protect its function too).

In this case, you can still have big problems with the following chain of events:

- cpp11 function `B` calls an R level callback (setting `should_unwind_protect = FALSE`)
- R level callback itself calls cpp11 function `A`
- Say that `A` sets up some complex C++ objects with destructors
- Then that `A` ends up calling `cpp11::stop()` or something that longjmps, but `should_unwind_protect = FALSE` so it never set up a `setjmp()`.
- The longjmp jumps all the way back to `B`'s `setjmp()`, _completely bypassing any destructors needed by A_

The big issue here is that `A()` and `B()` on their own can look very harmless, and like code that a package author would write without thinking twice about it.

I've come up with a reprex package to demonstrate this issue:
https://github.com/DavisVaughan/testcpp11unwind

---

A few options:
- Possibly the `BEGIN_CPP11` entry macro should always reset `should_unwind_protect = TRUE` (making sure it exists first). Then `A`'s call to `unwind_protect()` would still use `setjmp()` and `R_UnwindProtect()`.
- Consider removing the `cpp11_should_unwind_protect` global option altogether

---

The `cpp11_should_unwind_protect` nest guard is used for two things:

- For performance, i.e. you can wrap a tight loop where each iteration calls `unwind_protect()` in an outer `unwind_protect()` outside the loop so the protection is only set up once (mostly an issue with character vectors)
- For safety, i.e. the following code doesn't work without the nest guard

```cpp
[[cpp11::register]]
void test() {
  cpp11::unwind_protect([&] {
    cpp11::unwind_protect([&] {
      Rf_error("oh no!");
    });
  });
}
```

If `test()` is called from R:
- It goes through `.Call` and our wrapper, which sets up the `BEGIN_CPP11` and `END_CPP11` macros
- It calls `unwind_protect()` which calls `R_UnwindProtect()`, a C function!
- Inside `R_UnwindProtect()`, we call `unwind_protect()`
- From there we `Rf_error()`
- The inner `unwind_protect()` catches that C error and promotes it to a C++ exception which is `throw`n.
- That is thrown _across_ C stack frames, because we are inside the outer `R_UnwindProtect()` and no try/catch was set up within that to catch the unwind exception. The only try/catch is that most outer one set up by the macros
- That is UB and R crashes due to an uncaught exception

DavisVaughan:
Closes #325 
Closes #326 

- [x] Update documentation to talk about dangers of nested unwind protection
- [x] Update documentation to talk about dangers of using C++ code that calls destructors from within a manual `unwind_protect()` call (since the destructors won't run if a longjmp occurs)
- [x] Update documentation to talk about performance of cpp11 character vectors in a loop, and how to use `unwind_protect()` manually to avoid the performance hit
- [x] Update NEWS bullet
- [x] 2nd order revdep checking (these came back clean)

Note that this makes the example from https://github.com/r-lib/cpp11/issues/298 very slow again, because nested `unwind_protect()` is no longer "optimized". This is an example where character vector manipulation may be better with a manual `unwind_protect()` + R C API usage

### DIFF
--- a/vignettes/FAQ.Rmd
+++ b/vignettes/FAQ.Rmd
@@ -334,7 +334,7 @@ test_destructor_bad()
 #> Error: oh no!
 ```
 
-In general, the only code that can be called within `unwind_protect()` is "pure" C code. If you mix complex C++ objects with R's C API within `unwind_protect()`, then any R errors will result in a jump that prevents your destructors from running.
+In general, the only code that can be called within `unwind_protect()` is "pure" C code or C++ code that only uses POD (plain-old-data) types and no exceptions. If you mix complex C++ objects with R's C API within `unwind_protect()`, then any R errors will result in a jump that prevents your destructors from running.
 
 ##### Nested `unwind_protect()`
 
@@ -359,7 +359,7 @@ If you were to run `test_nested()` from R, it would likely crash or hang your R 
 - The outer `unwind_protect()` is called. It uses the C function `R_UnwindProtect()` to call its lambda function.
 - The inner `unwind_protect()` is called. It again uses `R_UnwindProtect()`, this time to call `Rf_error()`.
 - `Rf_error()` performs a `longjmp()` which is caught by the inner `unwind_protect()` and promoted to an exception.
-- That exception is thrown, but because we are in the outer call to `R_UnwindProtect()` (a C function), we end up throwing that exception _across_ C stack frames. This is _undefined behavior_, which typically results in that exception not being caught anywhere, causing C++ (and ultimately R) to abort.
+- That exception is thrown, but because we are in the outer call to `R_UnwindProtect()` (a C function), we end up throwing that exception _across_ C stack frames. This is _undefined behavior_, which is known to have caused R to crash on certain platforms.
 
 You might think that you'd never do this, but the same scenario can also occur with a combination of 1 call to `unwind_protect()` combined with usage of the cpp11 API:
 
@@ -418,7 +418,7 @@ If you've read the above bullet and still feel like you need to call `unwind_pro
 - You shouldn't use any parts of the cpp11 API that may call `unwind_protect()`.
 - You must be very careful not to call `unwind_protect()` in a nested manner.
 
-In other words, if you only use plain-old-data types and only use R's C API within the `unwind_protect()`, then you can use it.
+In other words, if you only use plain-old-data types, are careful to never throw exceptions, and only use R's C API, then you can use `unwind_protect()`.
 
 One place you may want to do this is when working with long character vectors. Unfortunately, due to the way cpp11 must protect the individual CHARSXP objects that make up a character vector, it can currently be quite slow to use the cpp11 API for this. Consider this example of extracting out individual elements with `x[i]` vs using the native R API:
 
--- a/vignettes/FAQ.Rmd
+++ b/vignettes/FAQ.Rmd
@@ -3,8 +3,8 @@ title: "FAQ"
 output: rmarkdown::html_vignette
 vignette: >
   %\VignetteIndexEntry{FAQ}
-  %\VignetteEngine{knitr::rmarkdown}
   %\VignetteEncoding{UTF-8}
+  %\VignetteEngine{knitr::rmarkdown}
 ---
 
 ```{r, include = FALSE}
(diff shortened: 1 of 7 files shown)


kevinushey (approved):
LGTM!

vignettes/FAQ.Rmd
```diff
@@ -283,3 +283,181 @@ add_one(x)
...
+
+#### 16. Ok but I really want to call `cpp11::unwind_protect()` manually
+
+If you've read the above bullet and still feel like you need to call `unwind_protect()`, then you should keep in mind the following when writing the function to unwind-protect:
+
+- You shouldn't create any C++ objects that have destructors.
+- You shouldn't use any parts of the cpp11 API that may call `unwind_protect()`.
+- You must be very careful not to call `unwind_protect()` in a nested manner.
+
+In other words, if you only use plain-old-data types and only use R's C API within the `unwind_protect()`, then you can use it.
```
lionel-, line 421: Also shouldn't throw exceptions (explicitly or implicitly).

lionel- (approved):
Nice writeup!

Issue #330, opened 2023-08-02
Issues with cpp11 0.4.4 changes to the preserve list structure


DavisVaughan:
The following will error if arrow is compiled with cpp11 0.4.4

```r
conn <- duckdb::dbConnect(duckdb::duckdb())

path <- gbifdb::gbif_example_data()
dataset <- arrow::open_dataset(path)

duckdb::dbDisconnect(conn, shutdown = TRUE)

# Now restart R
```

It should error on restart with:

```r
Error: bad value
```

This error is related to `SETCAR()` calls, which we confirmed with an lldb backtrace:

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001014589ec libR.dylib`Rf_error(format="bad value") at errors.c:961:26 [opt]
    frame #1: 0x00000001014bded4 libR.dylib`SETCAR(x=0x000000013000a0e0, y=0x00000001309d81c8) at memory.c:4285:2 [opt]
    frame #2: 0x0000000123188814 arrow.so`std::__1::__shared_ptr_emplace<cpp11::environment, std::__1::allocator<cpp11::environment> >::__on_zero_shared() [inlined] cpp11::$_0::release(this=<unavailable>, cell=<unavailable>) at protect.hpp:388:5 [opt]
    frame #3: 0x00000001231887d0 arrow.so`std::__1::__shared_ptr_emplace<cpp11::environment, std::__1::allocator<cpp11::environment> >::__on_zero_shared() [inlined] cpp11::sexp::~sexp(this=<unavailable>) at sexp.hpp:58:23 [opt]
    frame #4: 0x00000001231887cc arrow.so`std::__1::__shared_ptr_emplace<cpp11::environment, std::__1::allocator<cpp11::environment> >::__on_zero_shared() [inlined] cpp11::sexp::~sexp(this=<unavailable>) at sexp.hpp:58:11 [opt]
    frame #5: 0x00000001231887cc arrow.so`std::__1::__shared_ptr_emplace<cpp11::environment, std::__1::allocator<cpp11::environment> >::__on_zero_shared() [inlined] cpp11::environment::~environment(this=<unavailable>) at environment.hpp:21:7 [opt]
[...]

PR #331, opened 2023-08-02
Create 1 preserve list per compilation unit

DavisVaughan:
Closes https://github.com/r-lib/cpp11/issues/330 (see this issue for full discussion)
Closes https://github.com/r-lib/cpp11/issues/268 (as there is no longer a global option)

This is the minimal amount of work to move from a global protection list to 1 protection list per compilation unit.

In two follow up PRs I will:
- Remove `release_all()`, require R >= 3.5.0, remove `CPP11_USE_PRESERVE_OBJECT` and its related code (not much), remove the documentation section about R 3.5.0 and the tradeoffs about unwind-protect (#332)
- Talk to Lionel about how best to include the package name in a debug macro somehow, to make it easier to debug protection issues

### DIFF
--- a/inst/include/cpp11/protect.hpp
+++ b/inst/include/cpp11/protect.hpp
@@ -329,15 +329,15 @@ static struct {
   // This avoids issues with sharing preserve list state across compilation units
   // and across packages, which has historically caused many issues (#330).
   static SEXP get_preserve_list() {
-    static SEXP out = R_NilValue;
+    static SEXP out = init_preserve_list();
+    return out;
+  }
 
-    if (out == R_NilValue) {
-      // Initialize the list exactly once per compilation unit,
-      // and let R manage its memory
-      out = new_preserve_list();
-      R_PreserveObject(out);
-    }
-
+  static SEXP init_preserve_list() {
+    // Initialize the list exactly once per compilation unit,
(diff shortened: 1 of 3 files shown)


inst/include/cpp11/protect.hpp
```diff
@@ -350,58 +325,24 @@ static struct {
...
-    if (addr == nullptr) {
-      return R_NilValue;
-    }
-    return static_cast<SEXP>(addr);
-  }
+  // Each compilation unit purposefully gets its own preserve list.
+  // This avoids issues with sharing preserve list state across compilation units
+  // and across packages, which has historically caused many issues (#330).
+  static SEXP get_preserve_list() {
+    static SEXP out = R_NilValue;
```
lionel-, line 332: I think you can use dynamic init here?

lionel- (approved):
On further thought I think it'd still be nicer to store in the calling package's namespace. Not only for graph-domination purposes but also it'd make it possible to add some R-level tools in cpp11 to inspect all objects currently created by your cpp11 code, which might be useful in debugging situations.

It shouldn't be too hard to do that once we've added a symbol in the init file that represents the package namespace.

Not a blocker by any means, this PR is already good for now!

PR #337, opened 2023-10-23
Vectorize file argument to cpp_source()

krlmlr:
Use case: debugging if a particular phenomenon occurs if code is distributed across multiple source files.


pachadotdev:
@krlmlr @DavisVaughan I think this one should be merged, gcc 4.8 is vintage now

DavisVaughan:
Closed via #492 for a similar debugging use case

PR #340, opened 2023-11-08
Vendoring path

pachadotdev:
As @krlmlr suggested, it might be better to vendor to `src/vendor` instead of  `inst`

This PR implements that and also a better handling and messages about Makevars

The only error I can't see why it happens is

```
══ Results ══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 6.8 s

── Failed tests ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error (test-vendor.R:13:5): cpp_vendor: errors if cpp11 is already vendored
Error: cpp11 is not installed
Backtrace:
    ▆
 1. └─cpp11::cpp_vendor(pkg_path(pkg)) at test-vendor.R:13:5

Error (test-vendor.R:25:5): cpp_vendor: vendors cpp11
Error: cpp11 is not installed
Backtrace:
    ▆
 1. └─cpp11::cpp_vendor(pkg_path(pkg)) at test-vendor.R:25:5

[ FAIL 2 | WARN 0 | SKIP 0 | PASS 107 ]
```

PR #341, opened 2023-11-09
Vendoring

pachadotdev:
#340 but without the breaking changes


pachadotdev:
@krlmlr hi, this allows u to vendor to src/vendor!
@hadley i keep the unlink() function in vendor.R examples to pass the local check(), but that then creates problems with pkgdown

krlmlr:
Thanks. I'd wait for a comment on if we want to vendor into `src/vendor` at all. Perhaps there's a reason for vendoring into `inst/include` ?

pachadotdev:
> Thanks. I'd wait for a comment on if we want to vendor into `src/vendor` at all. Perhaps there's a reason for vendoring into `inst/include` ?

what i saw with check(), is that it doesn't like to put files in src, but if i vendor in src/vendor in a pkg project, it passes all checks for that another pkg

krlmlr:
- The checks in this PR are failing
- This PR seems to have a bit of noise
- We don't know the maintainers' perspective on if this PR is a good, so I'm not sure if it's worth addressing the first two issues at all

Let's not spend too much time here for now?

pachadotdev:
> * The checks in this PR are failing
> 
>     * This PR seems to have a bit of noise
> 
>     * We don't know the maintainers' perspective on if this PR is a good, so I'm not sure if it's worth addressing the first two issues at all
> 
> 
> Let's not spend too much time here for now?

sure, i was following from the vendoring in https://github.com/r-dbi/RPostgres/tree/main/src/vendor/boost

PR #345, opened 2023-11-28
Ensure format strings pass check in R-devel/Windows

paleolimbot:
Not sure exactly why this issue popped up on CI ( https://github.com/apache/arrow/pull/38894#issuecomment-1828488227 ), but it seems that at least one format string has `%x` (which expects an `int`) where I think the intention was to print memory addresses (i.e., `%p`). Our CI issue is probably because we're being a tad to aggressive about converting errors to warnings, but it seemed like an easy fix while I was taking a look!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # cpp11 (development version)
 
+* Internal changes requested by CRAN to fix invalid format string tokens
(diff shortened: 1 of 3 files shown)


DavisVaughan:
Thanks!

PR #349, opened 2023-12-18
Fix errors resulting from `-Wconversion -Wno-sign-conversion`

paleolimbot:
The M1 runner apparently runs checks with `-Wconversion -Wno-sign-conversion -Werror`, which has caused the arrow package to have a very long list of compiler warnings/build failures on the package check page ( https://www.stats.ox.ac.uk/pub/bdr/M1mac/arrow.log ). Most of these are in the arrow package (fixed by https://github.com/apache/arrow/pull/39250 ), but a few are in the cpp11 headers.

The biggest change is for `as_cpp<>()`, templates for which we instantiate in arrow for `int64_t`, and `uint8_t`, both of which interact poorly with various pieces of that logic. I'm happy to update this and add some tests but I wanted to check first to see if there's consensus on what should actually happen. A few cases that were exposed by these warnings are:

```cpp
as_cpp<int64_t>(NA_INTEGER); // static_cast<int64_t>(NA_INTEGER)?
as_cpp<int8_t>(NA_REAL); // error?
as_cpp<int8_t>(NA_INTEGER); // error?
as_cpp<int>(static_cast<double>(INT_MAX) + 1); // error?
```

### DIFF
--- a/inst/include/cpp11/as.hpp
+++ b/inst/include/cpp11/as.hpp
@@ -92,22 +92,21 @@ template <typename T>
 enable_if_integral<T, T> as_cpp(SEXP from) {
   if (Rf_isInteger(from)) {
     if (Rf_xlength(from) == 1) {
-      return INTEGER_ELT(from, 0);
+      return static_cast<T>(INTEGER_ELT(from, 0));
     }
   } else if (Rf_isReal(from)) {
     if (Rf_xlength(from) == 1) {
-      if (ISNA(REAL_ELT(from, 0))) {
-        return NA_INTEGER;
-      }
       double value = REAL_ELT(from, 0);
-      if (is_convertible_without_loss_to_integer(value)) {
-        return value;
+      if (ISNA(value) && sizeof(T) >= sizeof(int)) {
+        return static_cast<T>(NA_INTEGER);
+      } else if (!ISNA(value) && is_convertible_without_loss_to_integer(value)) {
+        return static_cast<T>(value);
       }
     }
   } else if (Rf_isLogical(from)) {
     if (Rf_xlength(from) == 1) {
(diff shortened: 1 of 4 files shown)


inst/include/cpp11/as.hpp
```diff
@@ -92,22 +92,21 @@ template <typename T>
...
-        return NA_INTEGER;
-      }
       double value = REAL_ELT(from, 0);
-      if (is_convertible_without_loss_to_integer(value)) {
-        return value;
+      if (ISNA(value) && sizeof(T) >= sizeof(int)) {
+        return static_cast<T>(NA_INTEGER);
+      } else if (!ISNA(value) && is_convertible_without_loss_to_integer(value)) {
+        return static_cast<T>(value);
       }
```
krlmlr, line 104: This gave me pause. Perhaps:

```suggestion
      if (ISNA(value)) {
        if (sizeof(T) >= sizeof(int)) {
          return static_cast<T>(NA_INTEGER);
        }
        // For T of smaller size, can't do better than throw an error
      } else if (is_convertible_without_loss_to_integer(value)) {
        return static_cast<T>(value);
      }
```

krlmlr (commented):
Thanks. Looks good, is it still necessary?

DavisVaughan (commented):
I've decided to close this unless CRAN turns these checks back on. Compiling cpp11tests locally with these warnings turned on makes a TON of mostly harmless warnings crop up. Like conversion of `size_t` to `R_xlen_t` without an explicit static_cast.

The `as_cpp()` one here is a bit more "real" but I think at some point we put the burden of using these on the user. Its tough to try and be fully generic and fully safe. It looks like there are other places in `as.hpp` where we'd also have to make similar changes, so I feel like its better to just keep it all as is for now rather than partially tweak some of these

PR #353, opened 2024-02-16
vendor path

pachadotdev:
A tidy (as in less noise and just 1 commit) version of #340

As @krlmlr suggested, it might be better to vendor to `src/vendor` instead of  `inst`

This PR implements that and also a better handling and messages about Makevars


tests/testthat/test-vendor.R
```diff
@@ -24,8 +24,8 @@ describe("cpp_vendor", {
 
     cpp_vendor(pkg_path(pkg))
```
krlmlr, line 25:
```suggestion
    cpp_vendor(file.path(pkg_path(pkg), "inst/include"))
```

Also add a test for vendoring into another directory?

R/vendor.R
```diff
@@ -9,13 +9,14 @@
 #' cpp11 currently installed on your machine.
 #'
 #' If you choose to vendor the headers you should _remove_ `LinkingTo:
-#' cpp11` from your DESCRIPTION.
+#' cpp11` from your DESCRIPTION. This is done automatically by this function.
```
krlmlr, line 12: Is this true?

R/vendor.R
```diff
@@ -48,16 +51,37 @@ cpp_vendor <- function(path = ".") {
...
-  files <- list.files(current, full.names = TRUE)
+  write_header(path, "cpp11.hpp", "cpp11", cpp11_header)
+
+  copy_files(list.files(current, full.names = TRUE), path, "cpp11", cpp11_header)
+
+  # Additional steps to make vendoring work ----
+
+  message(paste(
+    "Makevars and/or Makevars.win should have a line such as",
+    "'PKG_CPPFLAGS = -I../inst/include'"
```
krlmlr, line 62: Does this now need to show the new path?

krlmlr (commented):
@DavisVaughan: Any chance you'd consider vendoring into a different directory, perhaps by default? I think the current practice is somewhat dangerous because this makes cpp11 available for all dependants of the package it is vendored into. What's the reasoning behind choosing this path for vendoring?

DavisVaughan:
I don't currently have the resources to spend much time on cpp11, but in general I don't love vendoring in general because it doesn't _actually_ 100% insulate you from cpp11 changes. Previous versions of cpp11 relied on a global R option for the precious list, and that ends up getting shared between, say:
- Pkg A using vendored cpp11 0.4.5
- Pkg B using CRAN cpp11 0.4.7

So they aren't fully independent, and if there are changes made to the structure of the thing that gets stored in the global option (which has happened) then whoever sets up the global option first "wins" and can cause the other package to break

krlmlr:
Interesting. Do you think problems like this are going to resurface?

I've seen very weird behaviors when two cpp11 packages (one of them being duckdb) were used in the same R session. Could have been that.

For duckdb, I don't see a good alternative to vendoring. But if this isn't going to be widely supported, I'm fine with hacking my own solution to move from `inst/cpp11` .

R/vendor.R
pachadotdev: yes! I added the changes for that, because I am a dummy

R/vendor.R
pachadotdev: that line is added automatically, and uses the path the user wants

pachadotdev: i reallize i never pushed that change, done now

pachadotdev:
@krlmlr hi, I re-made the part that edits the DESCRIPTION

pachadotdev:
@krlmlr hi, I re-made the part that edits the DESCRIPTION

pachadotdev:
hi @DavisVaughan @krlmlr 
I made a small change because I sent another package to CRAN (cpp11armadillo), which was rejected for using `path = "."`
I have used `dir` and `subdir` now :)

pachadotdev:
btw, the only CI that fails is gcc 4.8 because of a bug with the compiler itself

Issue #356, opened 2024-05-15
build fails with gcc 4.8


pachadotdev:
The CI workflow passes the recent PRs except for gcc 4.8 on Ubuntu 20.

Here is what the CI says

```
/home/runner/work/_temp/Library/cpp11/include/cpp11/protect.hpp:161:31: warning: format not a string literal and no format arguments [-Wformat-security]
     decltype(std::declval<F*>()(std::declval<A&&>()...)) operator()(A&&... a) const {
                               ^
/home/runner/work/_temp/Library/cpp11/include/cpp11/protect.hpp: In instantiation of ‘struct cpp11::detail::closure<void(SEXPREC*, const char*, ...), SEXPREC*&, const char*&&>’:
/home/runner/work/_temp/Library/cpp11/include/cpp11/protect.hpp:185:89:   required from ‘void cpp11::protect::noreturn_function<F>::operator()(A&& ...) const [with A = {SEXPREC*&, const char*}; F = void(SEXPREC*, const char*, ...)]’
/home/runner/work/_temp/Library/cpp11/include/cpp11/protect.hpp:234:29:   required from ‘void cpp11::stop(const string&, Args ...) [with Args = {}; std::string = std::basic_string<char>]’
errors.cpp:5:84:   required from here
/home/runner/work/_temp/Library/cpp11/include/cpp11/protect.hpp:148:29: warning: format not a string literal and no format arguments [-Wformat-security]
   decltype(std::declval<F*>()(std::declval<Aref>()...)) operator()() && {
                             ^
‘
Internal compiler error: Error reporting routines re-entered.
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.8/README.Bugs> for instructions.
Preprocessed source stored into /tmp/ccOS2WLf.out file, please attach this to your bugreport.
make: *** [/opt/R/4.4.0/lib/R/etc/Makeconf:204: errors.o] Error 1
```

I tried to replicate this locally with VirtualBox, and so far I didn't succeed.

```
Enter one or more numbers, or an empty line to skip updates: 
Error: Could not find tools necessary to compile a package
Call `pkgbuild::check_build_tools(debug = TRUE)` to diagnose the problem.
```

DavisVaughan:
Isn't that extremely old? Like, "last release in 2015 and no longer supported" old? https://gcc.gnu.org/gcc-4.8/

pachadotdev:
[...]

PR #358, opened 2024-06-06
Remove non-API calls to R: SETLENGTH, SET_TRUELENGTH Fix #355

### LINKED ISSUE #355: Found non-API calls to R: `SETLENGTH`, `SET_TRUELENGTH`
Recently, my package `RcppAlgos` started getting NOTEs on several builds from CRAN checks:

https://cran.r-project.org/web/checks/check_results_RcppAlgos.html

The NOTE is: "Found non-API calls to R: `SETLENGTH`, `SET_TRUELENGTH`"

I  did many `grep` checks in my code repository and could not find and occurrence of `SETLENGTH` or `SET_TRUELENGTH`. I then googled the NOTE and several packages popped up. All of them that I checked depend on `cpp11`.

I then found this issue for `arrow`, apache/arrow#41404, which talks about this problem as well.

Both of these calls appear occurring in the `r_vector` header: https://github.com/r-lib/cpp11/blob/51f4cd5ad9425a491dedf951a3679346d416e51c/inst/include/cpp11/r_vector.hpp#L898-L903

I also found this from `r-source/wch` in `src/library/tools/R/sotools.R` (check out lines 561 to 669):

https://github.com/wch/r-source/blob/becbbf2a71ed56d4f7503d2dc4ac391d8c05eac6/src/library/tools/R/sotools.R#L561-L669

I'm not really sure what action to take here, I just wanted to bring this to your attention.

Regards,
Joseph

nx10:
Seems like R/CRAN does not want these used publicly anymore starting with 4.5.*

See details in #355

### DIFF
--- a/inst/include/cpp11/r_vector.hpp
+++ b/inst/include/cpp11/r_vector.hpp
@@ -896,14 +896,7 @@ inline void r_vector<T>::clear() {
 }
 
 inline SEXP truncate(SEXP x, R_xlen_t length, R_xlen_t capacity) {
-#if R_VERSION >= R_Version(3, 4, 0)
-  SETLENGTH(x, length);
-  SET_TRUELENGTH(x, capacity);
-  SET_GROWABLE_BIT(x);
-#else
-  x = safe[Rf_lengthgets](x, length);
-#endif
-  return x;
+  return safe[Rf_lengthgets](x, length);
 }
 
 template <typename T>


nx10:
For reference gcc 4.8 CI failing is unrelated: #356

chainsawriot:
@nx10 If you do it this way, you also have to modify this as well

https://github.com/r-lib/cpp11/blob/51f4cd5ad9425a491dedf951a3679346d416e51c/inst/include/cpp11/r_vector.hpp#L910-L926

Or else it will trigger `rchk` issues

ref ropensci/readODS#190

mtmorgan:
I see this issue in the [CRAN checks for rjsoncons](https://cran.r-project.org/web/checks/check_results_rjsoncons.html) too.

Reverting to `Rf_lengthgets()` as in the PR seems like a simple solution; should it be `Rf_xlengthgets()`, since elsewhere the code uses `Rf_xlength()`

```diff
diff --git a/inst/include/cpp11/r_vector.hpp b/inst/include/cpp11/r_vector.hpp
index 4831c2f..eee63f7 100644
--- a/inst/include/cpp11/r_vector.hpp
+++ b/inst/include/cpp11/r_vector.hpp
@@ -895,14 +895,8 @@ inline void r_vector<T>::clear() {
   length_ = 0;
 }
 
-inline SEXP truncate(SEXP x, R_xlen_t length, R_xlen_t capacity) {
-#if R_VERSION >= R_Version(3, 4, 0)
-  SETLENGTH(x, length);
-  SET_TRUELENGTH(x, capacity);
-  SET_GROWABLE_BIT(x);
-#else
-  x = safe[Rf_lengthgets](x, length);
-#endif
+inline SEXP truncate(SEXP x, R_xlen_t length) {
+  x = safe[Rf_xlengthgets](x, length);
   return x;
 }
 
@@ -914,11 +908,11 @@ inline r_vector<T>::operator SEXP() const {
     return data_;
   }
   if (length_ < capacity_) {
-    p->data_ = truncate(p->data_, length_, capacity_);
+    p->data_ = truncate(p->data_, length_);
     SEXP nms = names();
     auto nms_size = Rf_xlength(nms);
     if ((nms_size > 0) && (length_ < nms_size)) {
-      nms = truncate(nms, length_, capacity_);
+      nms = truncate(nms, length_);
       names() = nms;
     }
   }

```
@chainsawriot seems to suggest `PROTECT()`ing `names()`, which I guess satisfies rchk but doesn't seem necessary, `names()` is accessing an attribute of `data_`, and `data_` and its attributes are already preserved. On the other hand it looks like @jimhester originally [introduced `PROTECT()` in a number of places to satisfy rchk](https://github.com/r-lib/cpp11/commit/08733bd36b3419f896efc30218bdf6f925bf1cf0) even though probably not necessary.

`Rf_xlengthgets()` and `truncate()` "return a freshly allocated object" ([Writing R Extensions 5.9.2](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Allocating-storage)). Should the return value be `PROTECT()`ed? My C++ isn't strong enough to fully understand what the assignment is `p->data_ = truncate(...)`; if it's analogous to `p->data_ = sexp(truncate(...))` then no protection is necessary because `sexp()` immediately calls `preserved.insert()` which immediately protects it's argument. I am also not confident enough about what `names() = nms` does, and whether `nms` is immediately protected....

jimhester:
There will likely need to be benchmarking done to see how badly this effects performance. It was done this way to avoid an extra copy of the data when truncating. But additionally unwind protection used in `safe` also has some overhead when called in hot loops. 

As this function gets called when coercing cpp11 vectors to SEXP it is a pretty hot area of the code.

DavisVaughan:
We will be looking at this soon to get a patch release of cpp11 out. It seems that the R-devel non-API tweaks have slowed down to the point that we should be able to send a release in without much fear of something else additionally getting marked non-API later on.

In the meantime, if you have a package that uses cpp11 you can send in a release of that package even with the NOTE. CRAN is aware of it and will approve your package.

PR #365, opened 2024-07-27
Stop double protecting writable vectors

DavisVaughan:


cpp11test/src/test-doubles.cpp
```diff
@@ -233,7 +233,11 @@ context("doubles-C++") {
     w = z;
     expect_true(w.size() == 5);
     expect_true(w.data() != z.data());
-    expect_true(w.is_altrep() == z.is_altrep());
+    // Shallow duplication of objects of a very small size (like 1:5) don't result in
+    // a new ALTREP object. Make sure we check ALTREP-ness of the newly duplicated object,
+    // instead of just blindly inheriting the ALTREP-ness of the thing we duplicate.
+    expect_true(w.is_altrep() != z.is_altrep());
+    expect_true(w.is_altrep() == ALTREP(w.data()));
```
DavisVaughan, line 240: Found a bug here that resulted from us blindly copying over the `is_altrep_` field after performing a `Rf_shallow_duplicate()`, which can get rid of ALTREP-ness

cpp11test/src/test-integers.cpp
```diff
@@ -223,7 +223,6 @@ context("integers-C++") {
     R_xlen_t after = cpp11::detail::store::count();
 
     expect_true(before == 0);
-    // TODO: This should be 1 but writable vectors are being double protected
-    expect_true(after - before == 2);
+    expect_true(after - before == 1);
```
DavisVaughan, line 226: Yay!

inst/include/cpp11/doubles.hpp
```diff
@@ -84,7 +84,6 @@ template <>
 inline r_vector<double>::r_vector(std::initializer_list<named_arg> il)
     : cpp11::r_vector<double>(safe[Rf_allocVector](REALSXP, il.size())),
       capacity_(il.size()) {
-  protect_ = detail::store::insert(data_);
```
DavisVaughan, line 87: In all the specialization files, this was updating the writable `protect_`, which no longer exists

inst/include/cpp11/r_vector.hpp
```diff
@@ -78,30 +78,77 @@ class r_vector {
...
+    rhs.length_ = 0;
 
     return *this;
   };
 
+  // We are in read-only space so we can just copy over all properties except for
+  // `protect_`, which we need to manage on our own. `rhs` persists after this call, so we
+  // don't clear anything.
   r_vector(const r_vector& rhs) {
-    SEXP old_protect = protect_;
```
DavisVaughan, line 96: This is a constructor, so `protect_` starts out as `R_NilValue`. Nothing to do here, no need to try and release

inst/include/cpp11/r_vector.hpp
```diff
@@ -214,8 +261,6 @@ using has_begin_fun = std::decay<decltype(*begin(std::declval<T>()))>;
 template <typename T>
 class r_vector : public cpp11::r_vector<T> {
  private:
-  SEXP protect_ = R_NilValue;
```
DavisVaughan, line 217: Here is the important swap of this for `using cpp11::r_vector<T>::protect_;`

inst/include/cpp11/r_vector.hpp
```diff
@@ -224,6 +269,9 @@ class r_vector : public cpp11::r_vector<T> {
   using cpp11::r_vector<T>::data_p_;
   using cpp11::r_vector<T>::is_altrep_;
   using cpp11::r_vector<T>::length_;
+  using cpp11::r_vector<T>::protect_;
+
+  using cpp11::r_vector<T>::get_p;
 
   R_xlen_t capacity_ = 0;
```
DavisVaughan, line 276: This is now the only "new" field that writable vectors introduce

inst/include/cpp11/r_vector.hpp
```diff
@@ -302,8 +350,6 @@ class r_vector : public cpp11::r_vector<T> {
 
   explicit r_vector(const R_xlen_t size);
 
-  ~r_vector();
```
DavisVaughan, line 305: No longer need a destructor, which was just freeing the writable version of `protect_`.

The read only parent destructor is going to be called to free its version of `protect_` (which is now the only one)

inst/include/cpp11/r_vector.hpp
```diff
@@ -793,40 +827,40 @@ inline typename r_vector<T>::iterator r_vector<T>::find(const r_string& name) co
...
-      capacity_(rhs.capacity_) {}
+    : cpp11::r_vector<T>(safe[Rf_shallow_duplicate](rhs)), capacity_(rhs.capacity_) {}
 
 template <typename T>
 inline r_vector<T>::r_vector(r_vector<T>&& rhs)
-    : cpp11::r_vector<T>(rhs), protect_(rhs.protect_), capacity_(rhs.capacity_) {
-  rhs.data_ = R_NilValue;
-  rhs.protect_ = R_NilValue;
+    : cpp11::r_vector<T>(std::move(rhs)), capacity_(rhs.capacity_) {
+  rhs.capacity_ = 0;
```
DavisVaughan, line 835: This should still have the same optimization through the `std::move()`, which should force the move constructor of read only `r_vector` to get called (I'll double check this)

inst/include/cpp11/r_vector.hpp
```diff
@@ -793,40 +834,59 @@ inline typename r_vector<T>::iterator r_vector<T>::find(const r_string& name) co
...
 
-// We don't release the old object until the end in case we throw an exception
-// during the duplicate.
 template <typename T>
 inline r_vector<T>& r_vector<T>::operator=(const r_vector<T>& rhs) {
   if (data_ == rhs.data_) {
     return *this;
   }
 
-  cpp11::r_vector<T>::operator=(rhs);
```
DavisVaughan, line 821: This was a big part of where the double protection was accidentally coming into play before

Issue #360, opened 2024-07-24
BUG: cannot compile r-cpp11 when using MSVC's STL as C++ stdlib


h-vetinari:
In conda-forge we're building under some different constraints than for CRAN, which leads us to use different compilers etc.

As it turns out a [change](https://github.com/r-lib/cpp11/commit/28538e5cf45a588d1b44a88a704a23f190f31f3c#diff-c7ceacdb162dbfd5e85d5e1fdf7d0bdeab07929cebb0d8a7a8d9990f641fc72bR149) in r-cpp11 v0.4.4 ended up [breaking](https://github.com/apache/arrow/issues/43398) the R-bindings for arrow, only that we weren't able to root-cause the issue at the time, and so we haven't had windows builds for `r-arrow` since about 8 month.

After some digging together with the arrow folks, it seems to be a combination of two things
* the change to `std::transform` in https://github.com/r-lib/cpp11/commit/28538e5cf45a588d1b44a88a704a23f190f31f3c
* the fact that `cpp11::writable::r_vector<double>::iterator` has no copy assignment operator (implicitly deleted by the compiler)

It seems that the STL stumbles over this combination in its implementation of `std::transform`:
```
In file included from altrep.cpp:18:
In file included from .\./arrow_types.h:22:
In file included from .\.\./arrow_cpp11.h:20:
In file included from C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\include\memory:14:
In file included from C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\include\xmemory:15:
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\include\xutility:1408:13: error: object of type 'cpp11::writable::r_vector<double>::iterator' cannot be assigned because its copy assignment operator is implicitly deleted
 1408 |         _It = _STD forward<_UIter>(_UIt);
      |             ^
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\include\algorithm:3556:10: note: in instantiation of function template specialization 'std::_Seek_wrapped<cpp11::writable::r_vector<double>::iterator, cpp11::writable::r_vector<double>::iterator &>' requested here
 3556 |     _STD _Seek_wrapped(_Dest, _UDest);
      |          ^
[...]

Issue #334, opened 2023-08-28
std::max_element fails on writable vectors


thomasp85:
Using the `std::max_element()` algorithm on a writable vector results in a compile error. The error is not present when using a non-writable vector. It is speculated that the following commit made it possible on the standard vectors and need to be applied to the writable versions as well: https://github.com/r-lib/cpp11/commit/686ae04f4c239735bdbe3e462731ae650ec6a225

```c++
#include <cpp11/doubles.hpp>
[[cpp11::register]]
void test() {
  cpp11::writable::doubles foo = {1, 2, 3, 4, 5};
  std::max_element(foo.begin(), foo.end());
}
```

Compiling the above yields the following error:

```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/__algorithm/max_element.h:34:25: error: object of type 'cpp11::writable::r_vector<double>::iterator' cannot be assigned because its copy assignment operator is implicitly deleted
                __first = __i;
                        ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/__algorithm/max_element.h:44:19: note: in instantiation of function template specialization 'std::__max_element<std::__less<cpp11::writable::r_vector<double>::proxy> &, cpp11::writable::r_vector<double>::iterator>' requested here
    return _VSTD::__max_element<_Comp_ref>(__first, __last, __comp);
                  ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/__algorithm/max_element.h:52:19: note: in instantiation of function template specialization 'std::max_element<cpp11::writable::r_vector<double>::iterator, std::__less<cpp11::writable::r_vector<double>::proxy>>' requested here
    return _VSTD::max_element(__first, __last,
                  ^
code.cpp:4:8: note: in instantiation of function template specialization 'std::max_element<cpp11::writable::r_vector<double>::iterator>' requested here
  std::max_element(foo.begin(), foo.end());
       ^
/Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/library/cpp11/include/cpp11/r_vector.hpp:259:21: note: copy assignment operator of 'iterator' is implicitly deleted because field 'data_' is of reference type 'const cpp11::writable::r_vector<double> &'
    const r_vector& data_;
                    ^
1 error generated.
```

Whereas this compiles fine

```c++
void test(cpp11::doubles foo) {
[...]

PR #389, opened 2024-08-22
Enable copy assignment operator for `iterator` class

DavisVaughan:
Closes https://github.com/r-lib/cpp11/issues/360
Closes https://github.com/r-lib/cpp11/issues/334

Expands on https://github.com/r-lib/cpp11/commit/686ae04f4c239735bdbe3e462731ae650ec6a225 to carry over some of these changes to `iterator` as well (i.e., not just `const_iterator`). Note that I have not changed `iterator` to be a `random_access_iterator_tag`, I left it as a `forward_iterator_tag`. Making a random access iterator requires many extra methods. I don't even think we implement all of them for `const_iterator`. I've added some comments about that and if we have issues in the future with iterators we should look and see if we just need more methods.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # cpp11 (development version)
 
+* `cpp11::writable::r_vector<T>::iterator` no longer implicitly deletes its
+  copy assignment operator (#360).
+
+* `std::max_element()` can now be used with writable vectors (#334).
+
 * The `environment` class no longer uses the non-API function
   `Rf_findVarInFrame3()` (#367).
 
--- a/cpp11test/src/test-r_vector.cpp
+++ b/cpp11test/src/test-r_vector.cpp
@@ -4,6 +4,8 @@
 
 #include <testthat.h>
 
+#include <algorithm>  // for max_element
+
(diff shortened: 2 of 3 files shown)


DavisVaughan:
@h-vetinari I am expecting this to fix your conda issues!

h-vetinari:
Tested in https://github.com/conda-forge/r-arrow-feedstock/pull/94 using https://github.com/conda-forge/r-cpp11-feedstock/pull/27, and it works! 🥳 

Thank you very much! Looking forward to 0.4.8!

DavisVaughan:
Awesome!

h-vetinari:
> Thank you very much! Looking forward to 0.4.8!

Is there any timeline for this? :)

DavisVaughan:
It's on cran

h-vetinari:
Ah, I was looking at the git tags in repo here and didn't see anything past 0.4.7. Would still be good to tag the versions you release IMO. 🙃 

Thanks for the quick response!

DavisVaughan:
Woops i normally do. Added now https://github.com/r-lib/cpp11/releases/tag/v0.5.0

PR #396, opened 2024-08-26
Error if `get_namespace()` fails

### LINKED ISSUE #317: `package("<pkg>")` error if package namespace not found
Follow up to #316

DavisVaughan:
Closes #317

Previously, this would hang indefinitely

```r
cpp11::cpp_function('
SEXP test() {
  auto fn = cpp11::package("foo")["bar"];
  return fn();
}
')

test()
#> Error: object 'foo' not found
```

Now it at least says `foo` is not found, the same error that any other environment lookup failure would throw

### DIFF
--- a/inst/include/cpp11/function.hpp
+++ b/inst/include/cpp11/function.hpp
@@ -68,7 +68,7 @@ class package {
       return R_BaseEnv;
     }
     sexp name_sexp = safe[Rf_install](name);
-    return safe[Rf_findVarInFrame](R_NamespaceRegistry, name_sexp);
+    return safe[detail::r_env_get](R_NamespaceRegistry, name_sexp);
   }
 
   // Either base env or in namespace registry, so no protection needed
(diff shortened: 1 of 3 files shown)


inst/include/cpp11/function.hpp
```diff
@@ -68,7 +68,7 @@ class package {
       return R_BaseEnv;
     }
     sexp name_sexp = safe[Rf_install](name);
-    return safe[Rf_findVarInFrame](R_NamespaceRegistry, name_sexp);
+    return safe[detail::r_env_get](R_NamespaceRegistry, name_sexp);
```
DavisVaughan, line 71: I am confident about this switch:
- `r_env_get()` uses `Rf_findVarInFrame3(env, sym, TRUE)` on older R
- `Rf_findVarInFrame()` is exactly a wrapper around `Rf_findVarInFrame3(env, sym, TRUE)`

So `r_env_get()` is basically `Rf_findVarInFrame()` with a few extra checks - like erroring if the object is not found, as we want here.

On newer R it is using the "blessed" `R_getVar()` with `inherits = false`, which is also what we want.

PR #426, opened 2024-12-17
 as_integers and as_doubles now understand logical inputs

pachadotdev:
I added a test to verify that 1.0/0.0 (or 1/0) is the same as T/F.

The idea was to mimic the double to/from integer as method for logicals.


pachadotdev:
it would be good to merge [this](https://github.com/r-lib/cpp11/pull/405) PR and then adapt the tests here

PR #428, opened 2024-12-27
correctly set dimnames for matrices

### LINKED ISSUE #273: [Feature Request] colnames and rownames wrappers
Such as it's the case for `names()`, it would be extremely convenient to have `rownames()` and `colnames()` in cpp codes.

My idea (sorry that I don't enough C++) would be like:

```
// THIS WORKS
[[cpp11::register]]
doubles cpp_names(writable::doubles X) {
    X.attr("names") = {"a","b"};
    return X;
}

// THIS WON'T
// [[cpp11::register]]
// doubles_matrix<> cpp_colnames(writable::doubles_matrix<> X) {
//     writable::list dimnames;
//     dimnames.push_back(NULL);
//     dimnames.push_back({"a","b"});
//     X.attr("dimnames") = dimnames;
//     return X;
// }
```

pachadotdev:
There was a bug  in the matrix class such that attributes cannot be set (ref: https://stackoverflow.com/a/76019987/3720258). This was mentioned in #273.

With cpp11 0.5.1 I can add row and column names from C++ as long as the object is a `sexp` instead of a `*_matrix<>` (example: https://github.com/pachadotdev/economiccomplexity/blob/main/src/code.cpp#L28-L29)

With these changes, it is possible to assign colnames/rownames by borrowing from an input matrix or by setting `dimnames = list(...)`.

Acknowledgements to @stephematician! It only took me 2 yrs to take intensive C++ courses at UofT to 1st learn proper C++, and then use that knowledge for my own research.

~There is a problem with gh-actions. On my laptop, I can use `auto`, `attribute_proxy<V>` or `SEXP for lines 192-194 in matrix.hpp and it works. With GHA, it return an error.~ I fixed that with a workaround.


stephematician:
You can test on similar platforms to github actions with [`devtools::check_win_release`](https://devtools.r-lib.org/reference/check_win.html) and [`devtools::check_rhub`](https://devtools.r-lib.org/reference/check_rhub.html).

pachadotdev:
> You can test on similar platforms to github actions with [`devtools::check_win_release`](https://devtools.r-lib.org/reference/check_win.html) and [`devtools::check_rhub`](https://devtools.r-lib.org/reference/check_rhub.html).

yes! it's fixed now

stephematician:
Nicely done!

Not that I'm a cpp11 developer - but as a matter of style, I would make the tests 'simpler' and more clear about what they test. For example, there's no need to calculate logarithms. I would test two things specifically:

1.  That you can assign dimnames given two (suitable) vectors in a list; e.g. a function with signature `cpp11::doubles_matrix<> matrix_with_dimnames(int n, int m, cpp11::list dimnames) { /* ... */ }`
2.  That you can assign dimnames using another matrices' dimnames, e.g. a function like `cpp11::doubles_matrix<> duplicate_with_dimnames(cpp11::doubles_matrix<> x) { /* ... */ }`

pachadotdev:
> Nicely done!
> 
> Not that I'm a cpp11 developer - but as a matter of style, I would make the tests 'simpler' and more clear about what they test. For example, there's no need to calculate logarithms. I would test two things specifically:
> 
>     1. That you can assign dimnames given two (suitable) vectors in a list; e.g. a function with signature `cpp11::doubles_matrix<> matrix_with_dimnames(int n, int m, cpp11::list dimnames) { /* ... */ }`
> 
>     2. That you can assign dimnames using another matrices' dimnames, e.g. a function like `cpp11::doubles_matrix<> duplicate_with_dimnames(cpp11::doubles_matrix<> x) { /* ... */ }`

good point ! the log was just a "my brain defaults to log and OLS"

pachadotdev:
it would be good to merge [this](https://github.com/r-lib/cpp11/pull/405) PR and then adapt the tests here

PR #437, opened 2025-01-02
ordered and unordered C++ maps are converted to R lists

pachadotdev:
I did not want to clash with doubles.hpp and function.hpp, so probably there is a clever way to template my change


pachadotdev:
hi @DavisVaughan and @hadley
any change to merge this one?

PR #443, opened 2025-01-13
initial headers for dense Matrix (not matrix)

pachadotdev:
Just dgeMatrix and dsyMatrix for now.

### DIFF
--- a/cpp11test/R/cpp11.R
+++ b/cpp11test/R/cpp11.R
@@ -108,6 +108,18 @@ row_sums <- function(x) {
   .Call(`_cpp11test_row_sums`, x)
 }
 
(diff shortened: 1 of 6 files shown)


DavisVaughan:
Just FYI I am unlikely to add support for Matrix directly in cpp11 but it seems like a good candidate for an extension package.

pachadotdev:
> Just FYI I am unlikely to add support for Matrix directly in cpp11 but it seems like a good candidate for an extension package.

noted! I'll move it to extensions

pachadotdev:
@DavisVaughan could you check my other PRs? roxygen, complex numbers, complex matrices etc?

Issue #445, opened 2025-01-21
Another rchk issue to look into


DavisVaughan:
```
Function cpp11::writable::r_vector<SEXPREC*>::r_vector(std::initializer_list<cpp11::named_arg>)::{lambda()#1}::operator()() const
  [UP] unprotected variable names while calling allocating function Rf_mkCharCE cpp11/include/cpp11/list.hpp:91

Function cpp11::writable::r_vector<cpp11::r_bool>::r_vector(std::initializer_list<cpp11::named_arg>)::{lambda()#1}::operator()() const
  [UP] unprotected variable names while calling allocating function Rf_mkCharCE cpp11/include/cpp11/r_vector.hpp:890
```

https://raw.githubusercontent.com/kalibera/cran-checks/master/rchk/results/adfExplorer.out
https://raw.githubusercontent.com/kalibera/cran-checks/master/rchk/results/cpp11bigwig.out

I believe it is saying we have an unprotected SEXP called `names` at the point that we call `Rf_mkCharCE()`. This is surely a false alarm, because `Rf_setAttrib()` protects `names` by installed it as an attribute on `data_`, but I guess we can try to work around it.

```c++
template <>
inline r_vector<SEXP>::r_vector(std::initializer_list<named_arg> il)
    : cpp11::r_vector<SEXP>(safe[Rf_allocVector](VECSXP, il.size())),
      capacity_(il.size()) {
  unwind_protect([&] {
    SEXP names = Rf_allocVector(STRSXP, capacity_);
    Rf_setAttrib(data_, R_NamesSymbol, names);

    auto it = il.begin();

    for (R_xlen_t i = 0; i < capacity_; ++i, ++it) {
      SEXP elt = it->value();
      set_elt(data_, i, elt);

      SEXP name = Rf_mkCharCE(it->name(), CE_UTF8);
      SET_STRING_ELT(names, i, name);
    }
  });
}
```

pachadotdev:
@DavisVaughan how about https://github.com/r-lib/cpp11/pull/457 ?

kalibera:
That setAttrib() protects names during its own processing is another thing. What I meant is that after setAttrib() finishes, the argument "names" that has been passed to it may not be protected any longer as it would not be (by reference) used by the object. When names is a string vector and x is a vector, this wouldn't happen - it the names object will be re-used by reference. But it would happen e.g. for pairlists, see namesgets().

DavisVaughan:
Closed by #457

PR #457, opened 2025-05-09
implement fix for #445

pachadotdev:


pachadotdev:
:O I can't believe I got a commit merged into r-lib

Issue #453, opened 2025-04-26
Values added to a vector with `push_back` cannot be used immediately


pepijn-devries:
I'm using a `writable::integers` to store results in an iterative process, where I want to use elements from the vector of integers in the process and update them. However, when I `push_back()` a value to the vector, and then immediately check its value I will get some random number. Weirdly, when the vector is returned to R it seems to look fine. It is illustrated in the reprex below where I print the value that I just pushed back to the vector.

Am I doing something wrong?

``` r
library(cpp11)
#> Warning: package 'cpp11' was built under R version 4.3.3

cpp_function("
writable::integers test() {
    writable::integers x((R_xlen_t)0);
    for (int i = 0; i < 10; i++) {
      x.push_back(i);
      Rprintf(\"When I want to us the integer x[i], I get %i instead of %i\\n\", x[i], i);
    }
    return x;
  }
")

test()
#> When I want to us the integer x[i], I get -1107622224 instead of 0
#> When I want to us the integer x[i], I get -1107622224 instead of 1
#> When I want to us the integer x[i], I get -1107622224 instead of 2
#> When I want to us the integer x[i], I get -1107622224 instead of 3
#> When I want to us the integer x[i], I get -1107622224 instead of 4
#> When I want to us the integer x[i], I get -1107622224 instead of 5
#> When I want to us the integer x[i], I get -1107622224 instead of 6
#> When I want to us the integer x[i], I get -1107622224 instead of 7
#> When I want to us the integer x[i], I get -1107622224 instead of 8
#> When I want to us the integer x[i], I get -1107622224 instead of 9
#>  [1] 0 1 2 3 4 5 6 7 8 9
```

<sup>Created on 2025-04-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

pepijn-devries:
It turns out that I had to cast `x[i]` to `int` explicitly. After that correction the code above works as expected. Here is the corrected code:

```r
library(cpp11)

cpp_function("
writable::integers test() {
    writable::integers x((R_xlen_t)0);
    for (int i = 0; i < 10; i++) {
      x.push_back(i);
      Rprintf(\"When I want to us the integer x[i], I get %i which is the same as %i\\n\", (int)x[i], i);
    }
    return x;
  }
")

test()
```

Why is this explicit cast needed, and is this documented anywhere?

PR #464, opened 2025-08-26
Use values added to a vector with push_back immediately

pachadotdev:
implements #453 by delegating to the parent class


pachadotdev:
https://github.com/r-lib/cpp11/pull/463 got here somehow when creating a new branch :\

PR #481, opened 2026-01-15
Remove usage of `ATTRIB()`

jeroen:
This just merges [the change that Luke made](https://github.com/cran/cpp11/commit/6cc2e8bf673f4ccbcf08f64e3c203586239b5cf2), but conditional on the R version. 

This should fix the mess that CRAN created by publishing the package that only works on R-devel.


jeroen:
I don't understand this error.

Issue #295, opened 2022-10-17
`cpp11::warning(const char*)` crashes but `cpp11::warning(const std::string&)` does not


paleolimbot:
This is a very odd error and is difficult to reproduce; however, I haven't been able to isolate any change in Arrow's code that may have caused this to occur. The PR where this is observed is https://github.com/apache/arrow/pull/14271. Feel free to close this as obscure...I just thought it may be useful to document this somewhere in case anybody runs into this problem again.

Basically, on that one PR, all calls to `cpp11::warning(const char*, ...)` crash with a bad memory access in `std::tuple::get()`. `cpp11::warning(const std::string&, ...)` does not crash, nor does `cpp11::safe[R_warningcall](R_NilValue, const char*, ...)`. This is not the case anywhere except that PR; however, I don't need to invoke any changed code in that PR to make the crash happen (the PR is about ALTREP, but you can cause a crash with just a single function that only calls `cpp11::warning(const char*, ...)`

A reprex to illustrate the alternatives that work and the call that causes a crash:

``` r
# May or may not be important, but Arrow is on C++17
Sys.setenv("CXX_STD" = "CXX17")

cpp11::cpp_source(code = '

#include <cpp11.hpp>

// Not sure if this is related, but the crash always seems to be on
// tuple::get() and I wonder if two versions of this header are somehow
// finding their way into the objects
#include <tuple>

void my_safe_warningcall() {
  cpp11::safe[Rf_warningcall](R_NilValue, "some message");
}

void my_warning_string() {
  cpp11::warning(std::string("some message"));
}

void my_warning_const_char() {
  cpp11::warning("some message");
}


[[cpp11::register]]
void my_warning_call_wrapper(std::string opt) {
  if (opt == "safe_warningcall") {
    my_safe_warningcall();
  } else if (opt == "warning_string") {
    my_warning_string();
  } else {
    my_warning_const_char();
  }
}
')

# works!
my_warning_call_wrapper("safe_warningcall")
#> Warning: some message
# works!
my_warning_call_wrapper("warning_string")
#> Warning: some message
# crashes (on that one branch)!
my_warning_call_wrapper("warning_const_char")
#> Warning: some message
```

DavisVaughan:
I can't see anything that would be wrong here, so for now I'm going to close this as inactionable but we can come back to it if it bites us again

PR #493, opened 2026-05-05
Use tag types to generate attribute specific internals and avoid ODR violations

### LINKED ISSUE #491: Behavior change in `cpp11::warning(fmt, args...)` between 0.5.3 and 0.5.4 (clang only)
readxl built against **cpp11 0.5.4** segfaults at runtime when emitting a coercion warning. Built against **cpp11 0.5.3** (same readxl source, same R,same compiler), it works. Demo below.

In readxl:

* Issue: https://github.com/tidyverse/readxl/issues/784
* PR with workaround: https://github.com/tidyverse/readxl/pull/785
  - Replaces `cpp11::warning(fmt, args...)` / `cpp11::stop(fmt, args...)` with helpers that pre-format the message into a `std::string` — bypassing cpp11's variadic-template forwarding entirely.

Problem observed in various settings:
- r-universe (segfault; this was the first tipoff — see linked readxl issue).
- CRAN: `r-devel-linux-x86_64-{debian,fedora}-clang` (segfault); `r-release-macos-x86_64` (segfault);
  Additional checks on `M1Mac` (different presentation, but CRAN and I both think likely has same root cause). https://cran.r-project.org/web/checks/check_results_readxl.html. I also have an email from BDR and a May 15 deadline to fix.
- GHA: `macos-latest` (indefinite hang during vignette build, similar to CRAN M1Mac observation). Example: https://github.com/tidyverse/readxl/actions/runs/25193092291.
- Local: macOS arm64, Apple clang 21.

Not reproduced on gcc anywhere (CRAN's gcc flavors and Linux/Windows GHA all pass). Not R-version-specific: R 4.5 and R 4.6 both crash with cpp11 0.5.4.

Demo notes:
- R 4.5 on any clang platform. (cpp11 0.5.3 doesn't compile against R 4.6 — uses removed `R_NamespaceRegistry` — so R 4.5 is just where both versions build.)
- readxl pinned to v1.4.5 (current CRAN release, predates the workaround). cpp11 is header-only, so the buggy templates are baked into `readxl.so` at install time; rebuild from source for each cpp11 version.
- Build with `-g -O0` for a clean, immediate segfault. At default `-O2` the failure mode varies on macOS (indefinite hang). My `~/.R/Makevars`:

```
CFLAGS=-g -O0 -Wall
CXXFLAGS=-g -O0 -Wall
CXX11FLAGS=-g -O0 -Wall
CXX14FLAGS=-g -O0 -Wall
CXX17FLAGS=-g -O0 -Wall
F77=gfortran
FC=gfortran
```

Here are the two different setups. It's important to specify `?source?reinstall` in these pak calls to force recompilation and pick up the ambient cpp11.

```r
# --- cpp11 0.5.3 ---
pak::pkg_install("cpp11@0.5.3")
pak::pkg_install("tidyverse/readxl@v1.4.5?source&reinstall")

# --- cpp11 0.5.4 ---
pak::pkg_install("cpp11")  # 0.5.4 from CRAN
pak::pkg_install("tidyverse/readxl@v1.4.5?source&reinstall")
```

This tickles the problem because we'll see some coercion warnings from forcing things to become logical.

```r
library(readxl)

R.version.string
packageVersion("cpp11")
packageVersion("readxl")

read_excel(
  readxl_example("type-me.xlsx"),
  sheet = "logical_coercion",
  col_types = c("logical", "skip")
)
```

With cpp11 4.5.3, all is well. Notice the coercion warnings.

``` r
library(readxl)

R.version.string
#> [1] "R version 4.5.3 (2026-03-11)"
packageVersion("cpp11")
#> [1] '0.5.3'
packageVersion("readxl")
#> [1] '1.4.5'

read_excel(
  readxl_example("type-me.xlsx"),
  sheet = "logical_coercion",
  col_types = c("logical", "skip")
)
#> Warning: Expecting logical in A5 / R5C1: got a date
#> Warning: Expecting logical in A8 / R8C1: got 'cabbage'
#> # A tibble: 10 × 1
#>    `maybe boolean?`
#>    <lgl>
#>  1 NA
#>  2 FALSE
#>  3 TRUE
#>  4 NA
#>  5 TRUE
#>  6 FALSE
#>  7 NA
#>  8 TRUE
#>  9 FALSE
#> 10 FALSE
```

With cpp11 0.5.4, we see a segfault:

```r
> library(readxl)
>
> R.version.string
[1] "R version 4.5.3 (2026-03-11)"
> packageVersion("cpp11")
[1] '0.5.4'
> packageVersion("readxl")
[1] '1.4.5'
> read_excel(
+   readxl_example("type-me.xlsx"),
+   sheet = "logical_coercion",
+   col_types = c("logical", "skip")
+ )

 *** caught segfault ***
address 0xffff11000000, cause 'invalid permissions'

Traceback:
 1: read_fun(path = path, sheet_i = sheet, limits = limits, shim = shim,     col_names = col_names, col_types = col_types, na = na, trim_ws = trim_ws,     guess_max = guess_max, progress = progress)
 2: tibble::as_tibble(l, .name_repair = .name_repair)
 3: set_readxl_names(read_fun(path = path, sheet_i = sheet, limits = limits,     shim = shim, col_names = col_names, col_types = col_types,     na = na, trim_ws = trim_ws, guess_max = guess_max, progress = progress),     .name_repair = .name_repair)
 4: read_excel_(path = path, sheet = sheet, range = range, col_names = col_names,     col_types = col_types, na = na, trim_ws = trim_ws, skip = skip,     n_max = n_max, guess_max = guess_max, progress = progress,     .name_repair = .name_repair, format = format)
 5: read_excel(readxl_example("type-me.xlsx"), sheet = "logical_coercion",     col_types = c("logical", "skip"))
```

DavisVaughan:
Closes #491 
Closes #422 (because all this did was change the `warning()` signature to something that no longer conflicted with `stop()`, so it didn't fix the underlying problem)
This was the actual solution for #295 

My inline documentation fully describes the problem, so I will regurgitate it here

```
// Tag types to force templated `struct closure` and `apply()` infrastructure shared
// across `struct function` and `struct noreturn_function` to generate different
// attribute specific `struct closure` and `apply()` variants.
//
// Consider:
//
// ```
// cpp11::stop("error: %s", message)
// cpp11::warning("warning: %s", message)
// ```
//
// These both end up constructing the exact same templated `struct closure` and `apply()`
// functions. The `args` for the underlying `Rf_errorcall()` and `Rf_warningcall()` are:
// - `R_NilValue`
// - `const char* fmt`
// - `const char* message`
//
// The only difference is that `cpp11::stop()` is marked as `[[noreturn]]` because the
// underlying `Rf_errorcall()` is also marked as `[[noreturn]]` /
// `__attribute__((noreturn))`.
//
// But this causes issues! Due to C++'s ODR (One Definition Rule), only 1 variant of
// `apply()` and `struct closure` can be created per template combination. If the
// `cpp11::stop()` variant is linked in first, then some compilers use the `[[noreturn]]`
// hint on `cpp11::stop()` and `operator()` of `noreturn_function` to assert that the
// `apply()` function also cannot return, and returning is deemed unreachable. So then
// when `cpp11::warning()` tries to return from its call to `apply()`, a crash occurs. We
// see this output under ASAN: `execution reached an unreachable program point`.
//
// We've seen this issue on macOS and Linux under clang (gcc does not seem to reproduce
// this). To reproduce, you must have `cpp11::stop()` and `cpp11::warning()` calls in
// different translation units / files and the file containing `cpp11::stop()` must be
// linked first. Putting it first alphabetically seems to be enough, which is why we have
// `template-1-stop.cpp` and `template-2-warn.cpp` in our tests, along with
// `test-template.R` to test this exact issue. You also need to compile with `-O0`,
// otherwise you'll just get a hang rather than a crash.
//
// Adding the tag into the template definition forces `safe[fn]()` and
// `safe.noreturn[fn]()` calls to generate different `apply()` variants, avoiding this
// issue.
//
// https://github.com/r-lib/cpp11/issues/491
// https://github.com/r-lib/cpp11/issues/295
struct return_tag {};
struct no_return_tag {};
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # cpp11 (development version)
 
+* Fixed an issue where `cpp11::stop()` and `cpp11::warning()` calls with the same template instantiation could cause a crash on some systems (#491, #295).
+
 * `cpp_source()` now works with multiple `file`s (#492).
 
 # cpp11 0.5.4
--- a/cpp11test/DESCRIPTION
+++ b/cpp11test/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: cpp11test
 Title: A test suite and benchmark code for 'cpp11'
 Version: 0.0.0.9000
-Authors@R: 
+Authors@R:
     c(person(given = "Jim",
              family = "Hester",
              role = c("aut", "cre"),
@@ -21,3 +21,4 @@ Suggests:
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1
+Config/testthat/edition: 3
--- a/cpp11test/R/cpp11.R
+++ b/cpp11test/R/cpp11.R
@@ -236,6 +236,14 @@ rcpp_push_and_truncate_ <- function(size_sxp) {
   .Call(`_cpp11test_rcpp_push_and_truncate_`, size_sxp)
 }
 
+test_template_stop <- function() {
+  invisible(.Call(`_cpp11test_test_template_stop`))
+}
+
+test_template_warning <- function() {
+  invisible(.Call(`_cpp11test_test_template_warning`))
+}
+
 test_destruction_inner <- function() {
   invisible(.Call(`_cpp11test_test_destruction_inner`))
 }
--- a/cpp11test/src/cpp11.cpp
+++ b/cpp11test/src/cpp11.cpp
@@ -444,6 +444,22 @@ extern "C" SEXP _cpp11test_rcpp_push_and_truncate_(SEXP size_sxp) {
     return cpp11::as_sexp(rcpp_push_and_truncate_(cpp11::as_cpp<cpp11::decay_t<SEXP>>(size_sxp)));
   END_CPP11
 }
+// template-1-stop.cpp
+void test_template_stop();
+extern "C" SEXP _cpp11test_test_template_stop() {
+  BEGIN_CPP11
+    test_template_stop();
+    return R_NilValue;
+  END_CPP11
+}
+// template-2-warn.cpp
+void test_template_warning();
+extern "C" SEXP _cpp11test_test_template_warning() {
+  BEGIN_CPP11
+    test_template_warning();
+    return R_NilValue;
+  END_CPP11
+}
 // test-protect-nested.cpp
 void test_destruction_inner();
 extern "C" SEXP _cpp11test_test_destruction_inner() {
@@ -534,6 +550,8 @@ static const R_CallMethodDef CallEntries[] = {
     {"_cpp11test_sum_int_foreach_",         (DL_FUNC) &_cpp11test_sum_int_foreach_,         1},
     {"_cpp11test_test_destruction_inner",   (DL_FUNC) &_cpp11test_test_destruction_inner,   0},
     {"_cpp11test_test_destruction_outer",   (DL_FUNC) &_cpp11test_test_destruction_outer,   0},
+    {"_cpp11test_test_template_stop",       (DL_FUNC) &_cpp11test_test_template_stop,       0},
+    {"_cpp11test_test_template_warning",    (DL_FUNC) &_cpp11test_test_template_warning,    0},
     {"_cpp11test_upper_bound",              (DL_FUNC) &_cpp11test_upper_bound,              2},
     {"run_testthat_tests",                  (DL_FUNC) &run_testthat_tests,                  1},
     {NULL, NULL, 0}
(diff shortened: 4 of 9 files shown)


DavisVaughan:
Some additional prior issues / prs related to this

https://github.com/r-lib/cpp11/pull/85
https://github.com/r-lib/cpp11/issues/87

My theory is that these just didn't go far enough in making sure the internals were also attribute specific

DavisVaughan:
@jimhester - @jennybc and I thought you'd enjoy the resolution to this gnarly issue that has caused `cpp11::warning()` to crash for unexplainable reasons over the years

krlmlr:
Wonderful, thanks for looking into this, Davis!

